PR #3, opened 2018-08-10
Add Cocoapods podspec generation.

Lukasa:
Motivation:

A number of users have asked for the ability to install
NIOTransportServices via Cocoapods. This is an entirely reasonable
request, so let's do it.

Modifications:

Added a script that can generate, and optionally upload, a specfile
for NIOTransportServices.

Result:

Users can use cocoapods if they prefer.

### DIFF
--- a/scripts/build_podspec.sh
+++ b/scripts/build_podspec.sh
@@ -1,0 +1,81 @@
+#!/bin/bash
+##===----------------------------------------------------------------------===##
+##
+## This source file is part of the SwiftNIO open source project
+##
+## Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
+## Licensed under Apache License v2.0
+##
+## See LICENSE.txt for license information
+## See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+##
+## SPDX-License-Identifier: Apache-2.0
+##
+##===----------------------------------------------------------------------===##
+
+set -eu
+
+function usage() {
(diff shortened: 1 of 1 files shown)


scripts/build_podspec.sh
```diff
@@ -0,0 +1,81 @@
...
+  esac
+done
+shift "$((OPTIND-1))"
+
+if [[ $# -eq 0 ]]; then
+  echo "Must provide target version"
+  exit 1
+fi
+
+VERSION=$1
```
weissi: lower case

scripts/build_podspec.sh
```diff
@@ -0,0 +1,81 @@
...
+NAME="SwiftNIOTransportServices"
+
+here="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+tmpfile=$(mktemp -d /tmp/.build_podspecsXXXXXX)
```
weissi: that's not a file (maybe in the UNIX sense :P) so maybe `tmpdir`? Also we never delete it

Lukasa: Yeah the non-deletion is somewhat intentional, as it lets you view the output before you do anything with it. Open to changing that if you'd prefer.

weissi: na, its' fine

Lukasa:
@weissi mind re-reviewing?

PR #9, opened 2018-10-01
Fixes #8, and cleans some switch cases.

### LINKED ISSUE #8: NIOTSListener.bindPromise is disappearing
Looking at: https://github.com/apple/swift-nio-transport-services/blob/fb47fc1614ab391da593554cbf6718da90528f30/Sources/NIOTransportServices/NIOTSListenerChannel.swift#L282.

If we fail to create the `NWListener`, we call `self.close0()` with the error. Inside `close0` we immediately see that `self.nwListener` is `nil`, and exit doing nothing else. In this situation, `self.bindPromise`, aka the `promise` parameter to `beginActivating0()` is just left dangling.

I would guess that the catch block should look like this:

```swift
self.close0(error: error, mode: .all, promise: nil)
self.bindPromise?.fail(error: error)
self.bindPromise = nil
return
```

However, I'm not certain that's necessarily the right way to go about it. Any comments? I'm happy to submit the fix if my supposition above is actually correct.

AlanQuatermain:
Fixes #8, and cleans some switch cases.

### Motivation:

One bugfix, some cleanliness.

### Modifications:

`NIOTSListenerChannel.doClose0()` now evaluates and updates `self.bindPromise` irrespective of whether `self.nwListener` had been set.

Also, some case statements could be better: I've swapped `case _ as Thing:` with `case is Thing:`, and I've inlined the `let` value from `case _ as Thing: let thing = x as! Thing` to simply be `case let x as Thing:`.

### Result:

If the call to create the `NWListener` in `NIOTSListenerChannel.beginActivating0()` throws an error, any promise passed to that method will now correctly be updated, where before it would be silently ignored, potentially causing a waiting caller to wait forever.

Also, no case statements remain that might cause damage to my tooth enamel.

### DIFF
--- a/Sources/NIOTransportServices/NIOTSListenerChannel.swift
+++ b/Sources/NIOTransportServices/NIOTSListenerChannel.swift
@@ -147,14 +147,12 @@ extension NIOTSListenerChannel: Channel {
 
         // TODO: Many more channel options, both from NIO and Network.framework.
         switch option {
-        case _ as AutoReadOption:
+        case is AutoReadOption:
             // AutoRead is currently mandatory for TS listeners.
             if value as! AutoReadOption.OptionType == false {
                 throw ChannelError.operationUnsupported
             }
-        case _ as SocketOption:
-            let optionValue = option as! SocketOption
-
+        case let optionValue as SocketOption:
             // SO_REUSEADDR and SO_REUSEPORT are handled here.
             switch optionValue.value {
             case (SOL_SOCKET, SO_REUSEADDR):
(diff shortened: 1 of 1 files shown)


Sources/NIOTransportServices/NIOTSListenerChannel.swift
```diff
@@ -187,11 +185,9 @@ extension NIOTSListenerChannel: Channel {
         }
 
         switch option {
-        case _ as AutoReadOption:
+        case is AutoReadOption:
             return autoRead as! T.OptionType
-        case _ as SocketOption:
-            let optionValue = option as! SocketOption
-
+        case let optionValue as SocketOption:
```
Lukasa: These cleanups could be made in the upstream NIO repository as well: it might be nice if you could open a PR to do that.

PR #11, opened 2018-11-05
Support preconditionInEventLoop.

Lukasa:
Motivation:

In SwiftNIO 1.11 we'll ship improved interfaces for Channels to support
asserting being in the event loop. This is necessary because Dispatch
provides no non-precondition way to 100% guarantee that you are on a
specific queue, which is a requirement for accurate behaviour of
inEventLoop.

Right now this patch does not bring up the minimum required SwiftNIO version, so in some cases the old (more brittle) assertions will be used. We can raise the minimum required SwiftNIO version if we think it's worthwhile, but I suspect we don't actually need to do it: users who bump into this issue can just update SwiftNIO.

Modifications:

- Added an implementation of preconditionInEventLoop.

Result:

Accurate assertions

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -23,7 +23,7 @@ let package = Package(
         .executable(name: "NIOTSHTTPServer", targets: ["NIOTSHTTPServer"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/apple/swift-nio.git", from: "1.8.0"),
+        .package(url: "https://github.com/apple/swift-nio.git", from: "1.11.0"),
     ],
     targets: [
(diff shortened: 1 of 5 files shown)


Lukasa:
We don't *need* to: the situation is not any worse than it is today if we don't. I'm wondering if we gain much at all from doing it. That said, I'm happy to do it if you think it's worth it.

Lukasa:
Ok, this is updated to require SwiftNIO 1.11 and to elide internal uses of `inEventLoop` where they were intended for asserting.

PR #12, opened 2018-11-09
better diagnostics if unsupported ChannelOption set

weissi:
Motivation:

If the user sets an unsupported ChannelOption we should give the best
diagnostics possible.

Modifications:

Add the ChannelOption type to the error message that it's unsupported.

Result:

Happier users


normanmaurer:
late LGTM

PR #15, opened 2018-12-03
Support opting-out of Network.framework waiting state.

Lukasa:
Motivation:

In some cases users may prefer not to wait for Network.framework to
reattempt connection in the future. Users should be able to opt-out of the
default waiting behaviour in those cases.

Modifications:

- Added WaitForActivity ChannelOption.

Result:

Users can configure channels better.


Tests/NIOTransportServicesTests/NIOTSConnectionChannelTests.swift
```diff
@@ -487,4 +498,61 @@ class NIOTSConnectionChannelTests: XCTestCase {
             XCTFail("Unexpected error")
         }
     }
+
+    func testEarlyExitForWaitingChannel() throws {
+        let connectFuture = NIOTSConnectionBootstrap(group: self.group)
+            .channelOption(NIOTSChannelOptions.waitForActivity, value: false)
+            .connect(to: try SocketAddress(ipAddress: "127.0.0.1", port: 12345))
```
weissi: I'm not super keen on trying this on port 12345. Should we use a temp directory with `AF_UNIX` so we can 'guarantee' that it won't work?

Lukasa: Sure, why not.

Lukasa: Please re-review.

PR #16, opened 2018-12-06
Make Channel.isActive thread-safe.

Lukasa:
Motivation:

Sadly I overlooked the fact that Channel.isActive is supposed to be
safe to call from multiple threads: the implementation here was not.

Modifications:

Store the active state into an Atomic.

Result:

It will be thread-safe to ask if a channel is active.


Lukasa:
@weissi Done.

PR #17, opened 2018-12-06
Correctly hop across event loops during child channel setup.

Lukasa:
Motivation:

We were inadvertently modifying the registration information of
NIOTSEventLoop objects from multiple threads, which is very not good.

Modifications:

Add an event loop execute to hop loops.

Result:

Thread safe!

### DIFF
--- a/Sources/NIOTransportServices/NIOTSListenerChannel.swift
+++ b/Sources/NIOTransportServices/NIOTSListenerChannel.swift
@@ -339,11 +339,15 @@ extension NIOTSListenerChannel: StateManagedChannel {
     }
 
     public func channelRead0(_ data: NIOAny) {
+        self.eventLoop.assertInEventLoop()
+
         let channel = self.unwrapData(data, as: NIOTSConnectionChannel.self)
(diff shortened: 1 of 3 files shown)


Sources/NIOTransportServices/NIOTSListenerChannel.swift
```diff
@@ -339,11 +339,15 @@ extension NIOTSListenerChannel: StateManagedChannel {
     }
 
     public func channelRead0(_ data: NIOAny) {
+        self.eventLoop.assertInEventLoop()
+
         let channel = self.unwrapData(data, as: NIOTSConnectionChannel.self)
         let p: EventLoopPromise<Void> = self.eventLoop.newPromise()
```
weissi: shouldn't this be for `channel.eventLoop`?

Lukasa: As we use `channel.close` in the callback it doesn't *really* matter, but it's definitely faster, so let's do that.

Lukasa: Resolved.

PR #23, opened 2019-02-08
Bring patches up to new NIO 2 master

Lukasa:
Up to date with the new master.


Sources/NIOTransportServices/NIOTSConnectionBootstrap.swift
```diff
@@ -103,7 +103,7 @@ public final class NIOTSConnectionBootstrap {
     /// - returns: An `EventLoopFuture<Channel>` to deliver the `Channel` when connected.
     public func connect(host: String, port: Int) -> EventLoopFuture<Channel> {
         guard let actualPort = NWEndpoint.Port(rawValue: UInt16(port)) else {
-            return self.group.next().makeFailedFuture(error: NIOTSErrors.InvalidPort(port: port))
+            return self.group.next().makeFailedFuture(NIOTSErrors.InvalidPort(port: port))
```
weissi: this comment doesn't in this line but would it be worth to `import _NIO1APIShims` until we release 2.0.0 so that we don't constantly break this?

weissi: `InvalidPort` should also be `invalidPort` (lower case, right)?

weissi:
@swift-nio-bot test this please

PR #31, opened 2019-03-25
Correctly avoid branch dependency.

Lukasa:
Motivation:

When entering convergence we incorrectly specified a branch dependency
for something that is actually a tag.

Modifications:

- Remove branch dependency, use from: instead.

Result:

Users can use this repo!


weissi:
Oh whoops, sorry & thanks!

PR #33, opened 2019-03-26
update podspecs to NIO 2

weissi:
Motvation:

let's generate up to date podspecs

Modification:

update podspec

Result:

better podspec


weissi:
@Lukasa it even works, successfully built and run the NIOSMTP iOS app from: https://github.com/apple/swift-nio-examples/pull/10

PR #42, opened 2019-05-10
Annotate code with availability attributes

glbrntt:
Motivation:

It was not possible to import NIOTS into a project where
Network.framework was not supported by all deployment targets.

Modifications:

All NIOTS code, where applicable, was annotated with availability
attributes.

Result:

It is possible for application developers to import NIOTS on platforms
which do not support Network.framework without a compile time error.

### DIFF
--- a/Sources/NIOTransportServices/NIOTSEventLoop.swift
+++ b/Sources/NIOTransportServices/NIOTSEventLoop.swift
@@ -88,7 +88,7 @@ internal class NIOTSEventLoop: QoSEventLoop {
     }
 
     public init(qos: DispatchQoS) {
-    self.loop = DispatchQueue(label: "nio.transportservices.eventloop.loop", qos: qos, autoreleaseFrequency: .workItem)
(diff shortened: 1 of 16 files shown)


Sources/NIOTransportServices/NIOTSEventLoop.swift
```diff
@@ -85,7 +88,7 @@ internal class NIOTSEventLoop: QoSEventLoop {
     }
 
     public init(qos: DispatchQoS) {
-		self.loop = DispatchQueue(label: "nio.transportservices.eventloop.loop", qos: qos, autoreleaseFrequency: .workItem)
+    self.loop = DispatchQueue(label: "nio.transportservices.eventloop.loop", qos: qos, autoreleaseFrequency: .workItem)
```
Lukasa: This indentation is not more right than the one it replaced.

glbrntt: 🤦‍♂️

PR #45, opened 2019-05-22
Pass Channels down pipeline not NWConnection

### LINKED ISSUE #44: NIOTSListenerBootstrap incompatible with NIOExtras.ServerQuiescingHelper
### Expected behavior
I would like to use `NIOExtras.ServerQuiescingHelper` with `NIOTS` so I can better handle server quiescing. It works as intended with `NIO.ServerBootstrap` but not with `NIOTSListenerBootstrap `

### Actual behavior
The process crashes when processing a request:

```
Fatal error: tried to decode as type Channel but found NWConnection with contents other([C1 127.0.0.1:50360 tcp, interface: lo0, local: 127.0.0.1:50358, legacy-socket, server, path satisfied (Path is satisfied), interface: lo0, scoped]): file /Users/<truncated>/.build/checkouts/swift-nio.git--1696851023008440387/Sources/NIO/NIOAny.swift, line 200
```

### Steps to reproduce

The following code snippet illustrates how to reproduce the crash

1. Create a `NIOTS` HTTP server and start it
2. Invoke it (e.g. with `curl`)

### If possible, minimal yet complete reproducer code (or URL to code)

```swift
import NIO
import NIOTransportServices

let group = NIOTSEventLoopGroup()
let quiesce = ServerQuiescingHelper(group: group)
let serverChannel = NIOTSListenerBootstrap(group: group)
  .serverChannelInitializer { channel in
    // add the collection handler so all accepted child channels get collected
    channel.pipeline.add(handler: quiesce.makeServerChannelHandler(channel: channel))
  }
  .bind([...])
  .wait()

// make a request to the server, this will trigger the above crash.
```

### SwiftNIO version/commit hash

`NIO`: 2.1.1
`NIOTS`: 1.0.2
`NIOExtras`: 1.1.0

### Swift & OS version (output of `swift --version && uname -a`)

`Apple Swift version 5.0.1`

Lukasa:
Motivation:

The expectation is that server channels use Channel as their data type,
but the initial data type in NIOTSListenerChannel was actually
NWConnection. This is unnecessary and it makes it hard to interop
between NIOTS and NIO.

Modifications:

- Initialize the NIOTSConnectionChannel in the NIOTSListenerChannel
instead of in AcceptHandler.
- Added some missing @available annotations in the tests.

Result:

More consistency
Resolves #44.

### DIFF
--- a/Sources/NIOTransportServices/NIOTSListenerBootstrap.swift
+++ b/Sources/NIOTransportServices/NIOTSListenerBootstrap.swift
@@ -22,7 +22,7 @@ import Network
 @available(OSX 10.14, iOS 12.0, tvOS 12.0, *)
 public final class NIOTSListenerBootstrap {
     private let group: EventLoopGroup
-    private let childGroup: EventLoopGroup
+    private let childGroup: NIOTSEventLoopGroup
     private var serverChannelInit: ((Channel) -> EventLoopFuture<Void>)?
     private var childChannelInit: ((Channel) -> EventLoopFuture<Void>)?
     private var serverChannelOptions = ChannelOptionsStorage()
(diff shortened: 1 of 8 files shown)


Lukasa:
Pfft, fine. 😆

Lukasa:
Ok, test added that inserts something into the server channel initializer that expects a channel instead of an NWConnection.

PR #49, opened 2019-06-21
Lower EventLoopGroup requirements for creating bootstraps

glbrntt:
Motivation:

In NIO `ClientBootstrap` and `ServerBootstrap` are initialized with an
`EventLoopGroup`. Since `EventLoop` conforms to `EventLoopGroup` you can
initialize an bootstrap with an existing event loop. In NIO Transport
Services the bootstraps are initialized with a `NIOTSEventLoopGroup` and
as `NIOTSEventLoop` conforms to `EventLoop` and by extension
`EventLoopGroup` (but not `NIOTSEventLoopGroup`) it is not possible to
initialize a bootstrap with a pre-existing `NIOTSEventLoop`.

Modifications:

Change the bootstrap initializers to accept an `EventLoopGroup`.

Result:

NIO Transport Services bootstraps can be initialized with a
`NIOTSEventLoop`.

### DIFF
--- a/Sources/NIOTransportServices/NIOTSConnectionBootstrap.swift
+++ b/Sources/NIOTransportServices/NIOTSConnectionBootstrap.swift
@@ -21,7 +21,7 @@ import Network
 
 @available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
 public final class NIOTSConnectionBootstrap {
-    private let group: NIOTSEventLoopGroup
+    private let group: EventLoopGroup
     private var channelInitializer: ((Channel) -> EventLoopFuture<Void>)?
     private var connectTimeout: TimeAmount = TimeAmount.seconds(10)
     private var channelOptions = ChannelOptionsStorage()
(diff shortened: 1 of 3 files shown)


glbrntt:
cc @weissi

weissi:
looks SemVer minor to me because this compiles:

```
protocol EventLoopGroup {}

struct NIOTSEventLoop: EventLoopGroup {}

func foo(_ c: EventLoopGroup) {}

let f: (NIOTSEventLoop) -> Void = foo
```

but @glbrntt to verify using NIO's `check_no_api_breakages.sh`

glbrntt:
Not a fan either but I don't know of a better way to do this!

I also added comments to the initialisers taking `EventLoopGroup` to explain their existence.

weissi:
Thanks @glbrntt , having some explanation is great!

glbrntt:
Also I tested API breakage on a local branch which removed commit afbbead because of #50 -- I can check again once #51 is merged if you like?

Issue #50, opened 2019-07-03
master doesn't compile with 10.14 SDK


glbrntt:
`NWEndpoint.url` isn't in the 10.14 SDK.

glbrntt:
Fixed by #51

PR #51, opened 2019-07-03
Fix availability for NWEndpoint

glbrntt:
Motivation:

`NWEndpoint` added a `.url` case for macOS 10.15 which commit afbbead added
support for. However, in doing so it also broke support for 10.14.

Modifications:

Handle `.url` in the `default` case and also remove `@unknown` from the
cases of the switch statements in NIOTSListenerChannel and SocketAddress
init.

Result:

Compiles on 10.14.


glbrntt:
Fixes #50

glbrntt:
ping @Lukasa

PR #76, opened 2020-04-03
bootstraps: offer ELG validation

weissi:
Motivation:

Today, we just expect the ELGs passed to the bootstraps to be the
correct ones, if not, we crash.

Modifications:

Offer an alternative validatingGroup: init that just returns nil
if the ELGs are of the wrong types.

Result:

Easier to work with multi-stack systems for example when the user might
pass an ELG for either NIO on Sockets or NIO on Network.framework.

This is the NIOTS companion for https://github.com/apple/swift-nio/pull/1464

### DIFF
--- a/Sources/NIOTransportServices/NIOTSConnectionBootstrap.swift
+++ b/Sources/NIOTransportServices/NIOTSConnectionBootstrap.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the SwiftNIO open source project
 //
-// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
+// Copyright (c) 2017-2020 Apple Inc. and the SwiftNIO project authors
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information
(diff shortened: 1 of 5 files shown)


Sources/NIOTransportServices/NIOTSConnectionBootstrap.swift
```diff
@@ -17,6 +17,27 @@ import NIO
 import Dispatch
 import Network
 
+/// A `NIOTSConnectionBootstrap` is an easy way to bootstrap a `NIOTSConnectionChannel` when creating network clients.
+///
+/// Usually you re-use a `NIOTSConnectionBootstrap` once you set it up and called `connect` multiple times on it.
```
Lukasa, line 22:
```suggestion
/// Usually you re-use a `NIOTSConnectionBootstrap` once you set it up and call `connect` multiple times on it.
```

Sources/NIOTransportServices/NIOTSConnectionBootstrap.swift
```diff
@@ -30,20 +51,20 @@ public final class NIOTSConnectionBootstrap {
...
+    /// situations where it's impossible to tell ahead of time if the `EventLoopGroup` is compatible or not.
     ///
     /// - parameters:
     ///     - group: The `EventLoopGroup` to use.
-    public init(group: EventLoopGroup) {
-        self.group = group
+    public convenience init(group: EventLoopGroup) {
+        guard NIOTSBootstraps.isCompatible(group: group) else {
+            preconditionFailure("NIOTSConnectionBootstrap is only compatible with MultiThreadedEventLoopGroup and " +
+                                "SelectableEventLoop. You tried constructing one with \(group) which is incompatible.")
```
glbrntt, line 64: The compatible types need updating here

weissi: thanks! I did so much grepping but forgot that one :P

weissi: fixed

PR #82, opened 2020-04-23
Directly support the allowLocalEndpointReuse channel option

PeterAdams-A:
Directly support the allowLocalEndpointReuse channel option including mapping from shorthand options.  Also add shorthand options for client.

### Motivation:

The NetworkFramework directly supports the concept of allowLocalEndPointReuse.  Currently to get through the nio system, this is mapped to the SO_REUSEADDR option.  

Now there is a shorthand option for this, it is sad to map from allowLocalEndPointReuse to SO_REUSEADDR and back again.

### Modifications:

Add a new NIOTS channel option for this setting.
Set the underlying based on any of new setting, reuseAddr or reusePort.
Add a override to interpret the shorthand option directly into the new option.
Add shorthand options for client with tests of equivalence to long options.

### Result:

Behaviour is identical but we can feel happier that the option mapping is less confusing.

### DIFF
--- a/Sources/NIOTransportServices/NIOTSConnectionBootstrap.swift
+++ b/Sources/NIOTransportServices/NIOTSConnectionBootstrap.swift
@@ -250,7 +250,7 @@ extension NIOTSConnectionBootstrap: NIOClientTCPBootstrapProtocol {
     public func applyChannelOption(_ option: NIOTCPShorthandOption) -> Self? {
         // Currently the only thing we have special action for is allowImmediateEndpointAddressReuse
         // deal with that and let the default behaviour take care of everything else.
-        if option == NIOTCPShorthandOption.allowImmediateEndpointAddressReuse {
+        if option == NIOTCPShorthandOption.allowImmediateLocalEndpointAddressReuse {
             return channelOption(NIOTSChannelOptions.allowLocalEndpointReuse, value: true)
         }
         return .none
@@ -276,7 +276,7 @@ extension NIOTSConnectionBootstrap {
     @usableFromInline
     func channelOption(_ option: NIOTCPShorthandOption) -> NIOTSConnectionBootstrap {
         let applier = NIOTSConnectionBootstrap_Applier(contained: self)
-        return option.applyOptionDefaultMapping(with: applier).contained
+        return option.applyOptionDefaultMapping(applier).contained
     }
     
     fileprivate struct NIOTSConnectionBootstrap_Applier : NIOChannelOptionAppliable {
--- a/Sources/NIOTransportServices/NIOTSConnectionBootstrap.swift
+++ b/Sources/NIOTransportServices/NIOTSConnectionBootstrap.swift
@@ -247,7 +247,7 @@ extension NIOTSConnectionBootstrap: NIOClientTCPBootstrapProtocol {
     /// - parameters:
     ///     - options:  The options to try applying - the options applied should be consumed from here.
     /// - returns: The updated bootstrap with and options applied.
-    public func applyOptions(_ options: inout NIOTCPShorthandOptions) -> Self {
+    public func _applyOptions(_ options: inout NIOTCPShorthandOptions) -> Self {
         var toReturn = self
         if options.consumeAllowImmediateLocalEndpointAddressReuse().isSet {
             toReturn = channelOption(NIOTSChannelOptions.allowLocalEndpointReuse, value: true)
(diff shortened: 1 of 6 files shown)


PeterAdams-A:
Note - I've marked this as draft for now as it will require a new version of nio to exist containing https://github.com/apple/swift-nio/pull/1489 and having that set as a requirement in Package.swift here.

Sources/NIOTransportServices/NIOTSConnectionBootstrap.swift
```diff
@@ -242,5 +242,17 @@ public final class NIOTSConnectionBootstrap {
...
-extension NIOTSConnectionBootstrap: NIOClientTCPBootstrapProtocol {}
+extension NIOTSConnectionBootstrap: NIOClientTCPBootstrapProtocol {
+    /// Apply any understood shorthand options to the bootstrap, removing them from the set of options if they are consumed.
+    /// - parameters:
+    ///     - options:  The options to try applying - the options applied should be consumed from here.
+    /// - returns: The updated bootstrap with any understood options applied.
+    public func _applyChannelConvenienceOptions(_ options: inout ChannelOptions.TCPConvenienceOptions) -> Self {
+        var toReturn = self
+        if options.consumeAllowLocalEndpointReuse().isSet {
+            toReturn = channelOption(NIOTSChannelOptions.allowLocalEndpointReuse, value: true)
```
glbrntt, line 253:
```suggestion
            toReturn = self.channelOption(NIOTSChannelOptions.allowLocalEndpointReuse, value: true)
```

Tests/NIOTransportServicesTests/NIOTSBootstrapTests.swift
```diff
@@ -246,6 +246,72 @@ final class NIOTSBootstrapTests: XCTestCase {
...
+        let listenerChannel = try NIOTSListenerBootstrap(group: group)
+            .bind(host: "127.0.0.1", port: 0)
+            .wait()
+        
+        var optionValue : EventLoopFuture<Bool>? = nil
+        let bootstrap = NIOClientTCPBootstrap(NIOTSConnectionBootstrap(group: group),
+                                              tls: NIOInsecureNoTLS())
+            .channelConvenienceOptions([.allowLocalEndpointReuse])
+            .channelInitializer { channel in
+                optionValue = channel.getOption(NIOTSChannelOptions.allowLocalEndpointReuse)
```
glbrntt, line 261: Can't we just get the option from `client` once we've waited on the `connect` below? (and elsewhere in this PR)

Tests/NIOTransportServicesTests/NIOTSBootstrapTests.swift
```diff
@@ -246,6 +246,72 @@ final class NIOTSBootstrapTests: XCTestCase {
...
+            }
+        let client = try bootstrap.connect(to: listenerChannel.localAddress!).wait()
+        try client.close().wait()
+        
+        XCTAssertEqual(try optionValue!.wait(), true)
+    }
+    
+    func testShorthandOptionsAreEquvalent() throws {
+        func setAndGetOption<Option>(option: Option, _ applyOptions : (NIOClientTCPBootstrap) ->
+                NIOClientTCPBootstrap) throws -> Option.Value where Option : ChannelOption {
```
glbrntt, line 272: Breaking the closure signature on two lines looks _really_ weird. I suggest a new line after `Option,` and vertically aligning with the argument above:

```
func setAndGet(option: Option,
               foo: Bar) ...
```

Tests/NIOTransportServicesTests/NIOTSBootstrapTests.swift
```diff
@@ -246,6 +246,72 @@ final class NIOTSBootstrapTests: XCTestCase {
...
+                optionValue = channel.getOption(NIOTSChannelOptions.allowLocalEndpointReuse)
+                return channel.eventLoop.makeSucceededFuture(())
+            }
+        let client = try bootstrap.connect(to: listenerChannel.localAddress!).wait()
+        try client.close().wait()
+        
+        XCTAssertEqual(try optionValue!.wait(), true)
+    }
+    
+    func testShorthandOptionsAreEquvalent() throws {
```
glbrntt, line 270: typo "Equvalent"

Tests/NIOTransportServicesTests/NIOTSBootstrapTests.swift
PeterAdams-A: Yes, you're right - done.

Tests/NIOTransportServicesTests/NIOTSBootstrapTests.swift
PeterAdams-A: I have done that.  The return part of function feels a little odd now - toyed with various indentations but none looked great.

Tests/NIOTransportServicesTests/NIOTSBootstrapTests.swift
PeterAdams-A: Fixed.

PR #84, opened 2020-05-07
Update `NIOTSEventLoop` per apple/swift-nio#1508

gwynne:
Provides implementations of the new preconditions from apple/swift-nio#1508, tuned for use with a `DispatchQueue`-based `EventLoop`.

### Motivation:

See apple/swift-nio#1508 for rationale.

### Modifications:

Provides implementations of `preconditionInEventLoop(_:file:line:)` and `preconditionNotInEventLoop(_:file:line:)`.

### Result:

NIOTSEventLoop will now use dispatchPrecondition(condition:) for the in/not in preconditions instead of relying on the default implementation, yielding improved correctness.

**WARNING: As of this initial commit, the code contains multiple alternative implementations and considerable additional discussion in comments, describing ongoing concerns which make this draft entirely unsuitable for merging in its present form.**

### DIFF
--- a/Sources/NIOTransportServices/NIOTSEventLoop.swift
+++ b/Sources/NIOTransportServices/NIOTSEventLoop.swift
@@ -50,6 +50,30 @@ fileprivate enum LifecycleState {
     case closed
 }
 
+/// Reports a fatal error in the same manner as `preconditionFailure(_:file:line:)`. Requires Swift private API
+/// via `SwiftShims`. Used in `preconditionInEventLoop(_:file:line:)` and `preconditionNotInEventLoop(_:file:line:)`.
+/// This will almost definitely be removed before merging.
+#if !USE_PUBLIC_API_ONLY
+import SwiftShims
+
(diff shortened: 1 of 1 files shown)


Sources/NIOTransportServices/NIOTSEventLoop.swift
```diff
@@ -150,8 +174,89 @@ internal class NIOTSEventLoop: QoSEventLoop {
     }
 
     func preconditionInEventLoop(file: StaticString, line: UInt) {
+        // Do not forward to the other overload.
+        dispatchPrecondition(condition: .onQueue(self.loop))
+    }
+
+    func preconditionInEventLoop(_ message: @autoclosure() -> String, file: StaticString, line: UInt) {
```
Lukasa, line 181: Hmmmmm. I don't know that it's possible for us to meet the criteria for safely implementing this method correctly: specifically, we need to use only public API. I'm open to evolving it, but for now I think the safest thing to do is to simply discard the message. I know it's not good, but it's the only choice we have, and the absence of a debugging message, while highly annoying, is probably acceptable if the crash report is intact.

PR #87, opened 2020-05-13
Add supporter platforms to README

glbrntt:
Motivation:

We have a weird availability model for NIOTS since we can build it even
if Network.framework isn't available: we don't document this anywhere.

Modifications:

Explain supported platform model in the README.

Result:

Better documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,19 @@ and add the project as a sub-project by dragging it into your iOS project and ad
 
 Do note however that Network.framework requires macOS 10.14+, iOS 12+, or tvOS 12+.
 
+### Supported Platforms
+
+NIOTransportServices is supported where Network.framework is supported: macOS
+10.14+, iOS 12+, tvOS 12+, and watchOS 6+.
+
+In order to allow dependencies to use NIOTransportServices when it's available
+and fallback to NIO when it isn't, all code is behind import guards checking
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -43,6 +43,18 @@ and add the project as a sub-project by dragging it into your iOS project and ad
...
+### Supported Platforms
+
+NIOTransportServices is supported where Network.framework is supported: macOS
+10.14+, iOS 12+, and tvOS 12+.
+
+In order to allow dependencies to use NIOTransportServices when it's available
+and fallback to NIO when it isn't, all code is behind import guards checking
+the availability of Network.framework. As such NIOTransportServices may be
+built on platforms where Network.framework is *not* available.
+NIOTransportServices can be built on macOS 10.12+, iOS 10+, and tvOS 10+ but is
```
Lukasa, line 55: And Linux!

README.md
```diff
@@ -43,6 +43,18 @@ and add the project as a sub-project by dragging it into your iOS project and ad
...
+only functionally useful on macOS 10.14+, iOS 12+ and tvOS 12+.
```
Lukasa, line 56: We actually also support watchOS 6+ I think.

README.md
glbrntt: good point!

Issue #88, opened 2020-05-14
NWConnection and its NWParameters should be exposed for a channel


PasiSalenius:
### Expected behavior
This part in the Swift NIO Transport Services README made me assume that I could access the NWConnection and NWParameters of the established channel for a NIOTSListenerBootstrap or a NIOTSConnectionBootstrap.

> In addition to providing first-class support for Apple platforms, NIO Transport Services takes advantage of the richer API of Network.framework to provide more insight into the behaviour of the network than is normally available to NIO applications.

Basically, I'm looking to get statistics of a connection after its completion, things like NWParameters, NWProtocolMetadata, EstablishmentReports, DataTransferReports etc.

### Actual behavior
I cannot find any way to get to these connection details. The class NIOTSConnectionChannel seems to be internal, which prevents me from casting from Channel to that. Basically, all I'm left with is the same amount of connection info I had without NIO TS.

### If possible, minimal yet complete reproducer code (or URL to code)

```
let clientBootStrap = NIOTSConnectionBootstrap(group: context.eventLoop)
      .channelInitializer { channel in
            // no access to NIOTSConnectionChannel here...
      }
```

### SwiftNIO version/commit hash

swift-nio-transport-services version 1.5.1

### Swift & OS version (output of `swift --version && uname -a`)

Apple Swift version 5.2.2 (swiftlang-1103.0.32.6 clang-1103.0.32.51)

Lukasa:
Yeah, we have thus far not exposed top-level metadata like this. Generally speaking having access to the channel object would not help you either, as it has almost no public API surface area. What we need is some channel options that you can use to retrieve this information.

It seems like we could have a few channel options

- `currentPath`, for accessing the `NWConnection.currentPath`
- `metadata(definition)`, for accessing the metadata for a given `NWProtocol`.
- `establishmentReport`, for accessing the establishment report
- `dataTransferReport`, for initiating a data transfer report.

Are you interested in adding those channel options? We can provide some help so you know where to add them.

PasiSalenius:
Sure, I can give it a shot. It can take a few days before I get to it, though.

Lukasa:
Sure, no rush.

PasiSalenius:
Any suggestions how to best proceed getting the dataTransferReport out of the connection?

[...]

PR #90, opened 2020-05-31
New channel options exposing properties of underlying NWConnection

PasiSalenius:
This is a draft for implementing https://github.com/apple/swift-nio-transport-services/issues/88.

It adds new channel options in `NIOTSChannelOptions` and handles these in `getOption` of `NIOTSConnectionChannel`.

The collection of a `NWConnection.DataTransferReport` is not quite right in this implementation. I couldn't find a way to do it, so that part must probably be refactored. I posted what I have anyway, for sake of learning and discussion.

### DIFF
--- a/Sources/NIOTransportServices/NIOTSChannelOptions.swift
+++ b/Sources/NIOTransportServices/NIOTSChannelOptions.swift
@@ -25,16 +25,16 @@ public struct NIOTSChannelOptions {
     
     public static let currentPath = Types.NIOTSCurrentPathOption()
 
-    @available(OSX 10.15, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
-    public static let metadata = { (definition: NWProtocolDefinition) -> Types.NIOTSMetadataOption in
+    @available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 5.0, *)
+    public static let metadata = { (definition: NWProtocolDefinition) -> NIOTSChannelOptions.Types.NIOTSMetadataOption in
         .init(definition: definition)
     }
 
-    @available(OSX 10.15, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
-    public static let establishmentReport = Types.NIOTSEstablishmentReportOption()
+    @available(OSX 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
+    public static let establishmentReport = NIOTSChannelOptions.Types.NIOTSEstablishmentReportOption()
 
-    @available(OSX 10.15, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
-    public static let dataTransferReport = Types.NIOTSDataTransferReportOption()
+    @available(OSX 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
+    public static let dataTransferReport = NIOTSChannelOptions.Types.NIOTSDataTransferReportOption()
 }
 
 
@@ -82,7 +82,7 @@ extension NIOTSChannelOptions {
         /// `NIOTSMetadataOption` accesses the metadata for a given `NWProtocol`.
         ///
         /// This option is only valid with `NIOTSConnectionBootstrap`.
-        @available(OSX 10.15, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
+        @available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 5.0, *)
         public struct NIOTSMetadataOption: ChannelOption, Equatable {
             public typealias Value = NWProtocolMetadata
             
@@ -96,9 +96,9 @@ extension NIOTSChannelOptions {
         /// `NIOTSEstablishmentReportOption` accesses the `NWConnection.EstablishmentReport` of the underlying connection.
         ///
         /// This option is only valid with `NIOTSConnectionBootstrap`.
-        @available(OSX 10.15, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
+        @available(OSX 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
         public struct NIOTSEstablishmentReportOption: ChannelOption, Equatable {
-            public typealias Value = NWConnection.EstablishmentReport
+            public typealias Value = EventLoopFuture<NWConnection.EstablishmentReport?>
             
             public init() {}
         }
@@ -106,7 +106,7 @@ extension NIOTSChannelOptions {
         /// `NIOTSDataTransferReportOption` accesses the `NWConnection.DataTransferReport` of the underlying connection.
         ///
         /// This option is only valid with `NIOTSConnectionBootstrap`.
-        @available(OSX 10.15, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
+        @available(OSX 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
         public struct NIOTSDataTransferReportOption: ChannelOption, Equatable {
             public typealias Value = NWConnection.PendingDataTransferReport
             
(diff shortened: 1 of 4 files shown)


Sources/NIOTransportServices/NIOTSChannelOptions.swift
```diff
@@ -21,6 +22,19 @@ public struct NIOTSChannelOptions {
...
     public static let enablePeerToPeer = NIOTSChannelOptions.Types.NIOTSEnablePeerToPeerOption()
+    
+    public static let currentPath = Types.NIOTSCurrentPathOption()
+
+    @available(OSX 10.15, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
+    public static let metadata = { (definition: NWProtocolDefinition) -> Types.NIOTSMetadataOption in
+        .init(definition: definition)
+    }
+
+    @available(OSX 10.15, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
```
Lukasa, line 33: I think this guard (and associated guards) are wrong. Per [the documentation](https://developer.apple.com/documentation/network/nwconnection/3200452-requestestablishmentreport), the establishment report was added in macOS 10.15, iOS 13, tvOS 13, and watchOS 6.

Lukasa: The same applies to the data transfer report.

Sources/NIOTransportServices/NIOTSChannelOptions.swift
```diff
@@ -21,6 +22,19 @@ public struct NIOTSChannelOptions {
     public static let waitForActivity = NIOTSChannelOptions.Types.NIOTSWaitForActivityOption()
 
     public static let enablePeerToPeer = NIOTSChannelOptions.Types.NIOTSEnablePeerToPeerOption()
+    
+    public static let currentPath = Types.NIOTSCurrentPathOption()
+
+    @available(OSX 10.15, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
```
Lukasa, line 28: This guard (and associated guards) is wrong: the metadata accessor was available in macOS 10.14 and watchOS 5.

Sources/NIOTransportServices/NIOTSConnectionChannel.swift
```diff
@@ -349,6 +349,34 @@ extension NIOTSConnectionChannel: Channel {
...
+                    promise.succeed(pendingReport as! Option.Value)
+                } catch {
+                    promise.fail(error)
+                }
+                return promise.futureResult
+                
+            default:
+                break
+            }
+        }
```
Lukasa, line 378: These implementations should be in `getOptions0`: the current implementation is not thread-safe.

Sources/NIOTransportServices/NIOTSConnectionChannel.swift
```diff
@@ -388,7 +416,14 @@ extension NIOTSConnectionChannel: Channel {
             return self.options.waitForActivity as! Option.Value
         case is NIOTSChannelOptions.Types.NIOTSEnablePeerToPeerOption:
             return self.enablePeerToPeer as! Option.Value
+        case is NIOTSChannelOptions.Types.NIOTSCurrentPathOption:
+            return try self.nwConnection0().currentPath as! Option.Value
         default:
+            if #available(OSX 10.15, iOS 12.0, tvOS 12.0, watchOS 6.0, *) {
+                if let metadataOption = option as? NIOTSChannelOptions.Types.NIOTSMetadataOption {
```
Lukasa, line 423: Let's persist the `switch` pattern here.

Sources/NIOTransportServices/NIOTSConnectionChannel.swift
```diff
@@ -424,6 +459,13 @@ extension NIOTSConnectionChannel: StateManagedChannel {
             self = .open
         }
     }
+    
+    public func nwConnection0() throws -> NWConnection {
```
Lukasa, line 463: We can't add this, it's not thread-safe.

Sources/NIOTransportServices/NIOTSConnectionChannel.swift
PasiSalenius: As `requestEstablishmentReport` is asynchronous, how should I handle returning the value via a future? `getOption` returns a future, so that's why I placed the option reading there. `getOptions0` should return the value directly.

PasiSalenius:
I think I fixed the issues you pointed out.

I still can't figure out how to collect the dataTransferReport. I think I'm either finishing execution before the collecting completes with the closure, or I'm blocking my main thread with some waiting loop.

Sources/NIOTransportServices/NIOTSConnectionChannel.swift
Lukasa: In this case, the return value of the option should itself be a future. The futures returned from `getOption` are primarily used for dealing with the fact that that interface needs to be thread-safe: it's fine to return a nested future here.

Sources/NIOTransportServices/NIOTSChannelOptions.swift
```diff
@@ -21,6 +22,19 @@ public struct NIOTSChannelOptions {
     public static let waitForActivity = NIOTSChannelOptions.Types.NIOTSWaitForActivityOption()
 
     public static let enablePeerToPeer = NIOTSChannelOptions.Types.NIOTSEnablePeerToPeerOption()
+    
+    public static let currentPath = NIOTSChannelOptions.Types.NIOTSCurrentPathOption()
+
+    @available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 5.0, *)
```
Lukasa, line 28: Nit: we don't need this annotation, the one on `NIOTSChannelOptions` is sufficient (it's a strict superset of this one).

Sources/NIOTransportServices/NIOTSConnectionChannel.swift
```diff
@@ -388,13 +388,44 @@ extension NIOTSConnectionChannel: Channel {
...
+                    break
+                }
+            }
+            
             fatalError("option \(type(of: option)).\(option) not supported")
         }
     }
 }
 
-
```
Lukasa, line 397: Nit: remove this change.

Sources/NIOTransportServices/NIOTSConnectionChannel.swift
```diff
@@ -424,7 +455,7 @@ extension NIOTSConnectionChannel: StateManagedChannel {
             self = .open
         }
     }
-
+
```
Lukasa, line 458: Nit: 
```suggestion

```

Tests/NIOTransportServicesTests/NIOTSChannelOptionsTests.swift
```diff
@@ -0,0 +1,115 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
```
Lukasa, line 5:
```suggestion
// Copyright (c) 2020 Apple Inc. and the SwiftNIO project authors
```

Tests/NIOTransportServicesTests/NIOTSChannelOptionsTests.swift
```diff
@@ -0,0 +1,115 @@
...
+        let connection = try NIOTSConnectionBootstrap(group: self.group)
+            .connect(to: listener.localAddress!)
+            .wait()
+        defer {
+            XCTAssertNoThrow(try connection.close().wait())
+        }
+        
+        let pendingReport = try connection.getOption(NIOTSChannelOptions.dataTransferReport).wait()
+        
+        pendingReport.collect(queue: .main) { report in
```
Lukasa, line 109: Let's try collecting this on a new one-off serial queue and using `DispatchGroup` to wait for the result. Right now we don't wait for this to complete, but doing that should work well.

PasiSalenius:
Okay I think I got it. The test now passes with one path in the data transfer report. Also fixed all the nits.

PR #95, opened 2020-06-15
Add user event fired when waiting for connectivity.

Lukasa:
Motivation:

On Apple's devices it is considered a best practice to wait for network
connectivity in cases when a connection request cannot immediately be
satisfied, rather than erroring out. This reflects the dynamic and fluid
network environment on Apple devices, with their many network interfaces
and complex interactions between radios, VPNs, and network devices.

While NIOTS supports this model of interaction (and indeed uses it out
of the box), and supports configuring it (by setting
`NIOTSChannelOptions.waitForActivity`), a key pillar is missing:
observability. Right now we don't actually _tell_ the user when we're
waiting for connectivity. This makes it difficult to act on this
information.

Modifications:

- Added a user event, `NIOTSNetworkEvents.WaitingForConnectivity` that
  is fired when connectivity could not be established, but the situation
  may change in future.

Result:

Our users can tell their users when they're waiting for something!

### DIFF
--- a/Tests/NIOTransportServicesTests/NIOTSConnectionChannelTests.swift
+++ b/Tests/NIOTransportServicesTests/NIOTSConnectionChannelTests.swift
@@ -89,7 +89,7 @@ final class DisableWaitingAfterConnect: ChannelOutboundHandler {
 
 final class PromiseOnActiveHandler: ChannelInboundHandler {
     typealias InboundIn = Any
-    typealias InboudOut = Any
+    typealias InboundOut = Any
 
     private let promise: EventLoopPromise<Void>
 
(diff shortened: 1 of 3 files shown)


Tests/NIOTransportServicesTests/NIOTSConnectionChannelTests.swift
```diff
@@ -102,6 +102,27 @@ final class PromiseOnActiveHandler: ChannelInboundHandler {
...
+    private var eventWaiter: EventLoopPromise<Event>?
+
+    init(_ eventWaiter: EventLoopPromise<Event>) {
+        self.eventWaiter = eventWaiter
+    }
+
+    func userInboundEventTriggered(context: ChannelHandlerContext, event: Any) {
+        if let event = event as? Event {
+            var promise = Optional<EventLoopPromise<Event>>.none
+            swap(&promise, &self.eventWaiter)
```
glbrntt, line 119: This doesn't feel very idiomatic, is this working around something?

Lukasa: It's an alternative spelling for:

```swift
var promise = self.eventWaiter
self.eventWaiter = nil
promise?.succeed(event)
```

The argument is that we want to succeed this promise exactly once and then nil it out, so we do that.

We don't _have_ to do it here, but it's a habit I've gotten in to over time.

glbrntt: Fair enough, it just looked like a curious spelling to me!

Issue #79, opened 2020-04-03
Linux CI


weissi:
We need Linux CI here, sounds silly but #77  and #78 :'(. Just a Swift 5.0 build will do.

glbrntt:
hey @tomerd any update on this?

PeterAdams-A:
This is now done.

PR #99, opened 2020-07-23
Add docker config for swift-nio-transport-services

PeterAdams-A:
Motivation:

It's important that transport-services compiles and does nothing
on linux so it can be used in builds which just do the right thing
on all platforms.

Modifications:

Add basic docker-compose files based on those from swift-nio

Result:

It's possible to run docker on this repo to run the tests.

### DIFF
--- a/docker/docker-compose.1804.50.yaml
+++ b/docker/docker-compose.1804.50.yaml
@@ -1,0 +1,16 @@
+version: "3"
+
+services:
+
+  runtime-setup:
+    image: swift-nio-transport-services:18.04-5.0
+    build:
+      args:
+        ubuntu_version: "bionic"
+        swift_version: "5.0"
+
+  test:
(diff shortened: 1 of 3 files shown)


PeterAdams-A:
This feels like the first step towards solving issue  #79

docker/docker-compose.1804.50.yaml
```diff
@@ -0,0 +1,16 @@
...
+    build:
+      args:
+        ubuntu_version: "bionic"
+        swift_version: "5.0"
+
+  test:
+    image: swift-nio-transport-services:18.04-5.0
+
+  shell:
+    image: swift-nio:18.04-5.0
```
tomerd, line 16: swift-nio-transport-services

PeterAdams-A: Thanks, I'll fix in another PR.

PR #108, opened 2020-10-25
Remove agressive precondition in zero-length-writes.

Lukasa:
Motivation:

The zero-length writes handler will crash if it receives channelInactive
before channelActive. Sadly, that's totally possible if a channel is
closed from the connect promise.

Modifications:

- Tone back the preconditions a bit.
- Add a test that triggers this path.

Result:

Channel handler behaves better on early close

### DIFF
--- a/Sources/NIOTransportServices/NIOFilterEmptyWritesHandler.swift
+++ b/Sources/NIOTransportServices/NIOFilterEmptyWritesHandler.swift
@@ -88,9 +88,16 @@ public final class NIOFilterEmptyWritesHandler: ChannelDuplexHandler {
 // Connection state management
 extension NIOFilterEmptyWritesHandler {
     public func channelActive(context: ChannelHandlerContext) {
-        assert(self.state == .notActiveYet)
-        self.state = .open
-        context.fireChannelActive()
+        switch self.state {
+        case .notActiveYet:
+            self.state = .open
+            context.fireChannelActive()
+        case .closedFromLocal:
+            // Closed before we activated, not a problem.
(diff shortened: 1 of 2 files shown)


glbrntt (approved):
This LGTM. 

For my own understanding: do we only expect `channelInactive` before `channelActive` if closing locally?

Sources/NIOTransportServices/NIOFilterEmptyWritesHandler.swift
```diff
@@ -88,9 +88,16 @@ public final class NIOFilterEmptyWritesHandler: ChannelDuplexHandler {
...
-        context.fireChannelActive()
+        switch self.state {
+        case .notActiveYet:
+            self.state = .open
+            context.fireChannelActive()
+        case .closedFromLocal:
+            // Closed before we activated, not a problem.
+            assert(self.prefixEmptyWritePromise == nil)
+        case .open, .closedFromRemote, .error:
+            preconditionFailure()
```
Davidde94, line 99: Not worth having a tiny little message here?

Davidde94 (approved):

Sources/NIOTransportServices/NIOFilterEmptyWritesHandler.swift
Lukasa: I think it's largely unnecessary: on iOS crash reports will point directly at this line anyway, and the static string raises binary sizes with minimal improvement in debuggability.

Lukasa:
> For my own understanding: do we only expect `channelInactive` before `channelActive` if closing locally?

Yes.

PR #112, opened 2021-01-27
Tolerate out-of-band ports.

Lukasa:
Motivation:

We shouldn't crash if the user gives us a bad port.

Modifications:

- Check the ports are in-band before we move on.

Result:

We won't crash.


weissi (approved):
awesome, ta

PR #117, opened 2021-03-15
Add support for syncOptions to NIOTSListenerChannel and NIOTSConnectionChannel

glbrntt:
Motivation:

NIO 2.27.0 added optional support for synchronous channel options.
NIOTSListenerChannel and NIOTSConnectionChannel should support this.

Modifications:

- Add synchronous options for NIOTSConnectionChannel and
  NIOTSListenerChannel
- Avoid an allocation in 'getOption' for each channel if the caller is
  already on the right event loop by using 'makeSucceededFuture'
- Remove a no longer used internal function and an already dead private
  helper

Result:

- Support for sync options and fewer allocations

### DIFF
--- a/Sources/NIOTransportServices/NIOTSConnectionChannel.swift
+++ b/Sources/NIOTransportServices/NIOTSConnectionChannel.swift
@@ -22,31 +22,6 @@ import Dispatch
 import Network
 import Security
 
-/// Execute the given function and synchronously complete the given `EventLoopPromise` (if not `nil`).
-@available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
-func executeAndComplete<T>(_ promise: EventLoopPromise<T>?, _ body: () throws -> T) {
-    do {
-        let result = try body()
-        promise?.succeed(result)
-    } catch let e {
-        promise?.fail(e)
-    }
-}
-
-/// Merge two possible promises together such that firing the result will fire both.
-@available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
-private func mergePromises(_ first: EventLoopPromise<Void>?, _ second: EventLoopPromise<Void>?) -> EventLoopPromise<Void>? {
-    if let first = first {
-        if let second = second {
-            first.futureResult.cascade(to: second)
-        }
-        return first
-    } else {
-        return second
-    }
-}
-
-
 /// Channel options for the connection channel.
 @available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
 private struct ConnectionChannelOptions {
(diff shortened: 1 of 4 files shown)


Sources/NIOTransportServices/NIOTSConnectionChannel.swift
```diff
@@ -891,4 +862,30 @@ fileprivate extension ChannelState where ActiveSubstate == NIOTSConnectionChanne
...
+            try self.channel.setOption0(option: option, value: value)
+        }
+
+        @inlinable
+        public func getOption<Option: ChannelOption>(_ option: Option) throws -> Option.Value {
+            return try self.channel.getOption0(option: option)
+        }
+    }
+
+    public final var syncOptions: NIOSynchronousChannelOptions? {
```
Lukasa, line 887: I don't think this needs a `final` annotation, the class is already `final`.

Sources/NIOTransportServices/NIOTSConnectionChannel.swift
```diff
@@ -891,4 +862,30 @@ fileprivate extension ChannelState where ActiveSubstate == NIOTSConnectionChanne
...
+        fileprivate init(channel: NIOTSConnectionChannel) {
+            self.channel = channel
+        }
+
+        @inlinable
+        public func setOption<Option: ChannelOption>(_ option: Option, value: Option.Value) throws {
+            try self.channel.setOption0(option: option, value: value)
+        }
+
+        @inlinable
```
Lukasa, line 881: I think you can remove these inlinable definitions: this type is only accessed through an existential container so it can't be inlined through.

Sources/NIOTransportServices/NIOTSListenerChannel.swift
```diff
@@ -474,4 +470,30 @@ extension NIOTSListenerChannel {
...
+        fileprivate init(channel: NIOTSListenerChannel) {
+            self.channel = channel
+        }
+
+        @inlinable
+        public func setOption<Option: ChannelOption>(_ option: Option, value: Option.Value) throws {
+            try self.channel.setOption0(option: option, value: value)
+        }
+
+        @inlinable
```
Lukasa, line 489: Same note about inlinable.

Sources/NIOTransportServices/NIOTSListenerChannel.swift
```diff
@@ -474,4 +470,30 @@ extension NIOTSListenerChannel {
...
+        public func getOption<Option: ChannelOption>(_ option: Option) throws -> Option.Value {
+            return try self.channel.getOption0(option: option)
+        }
+    }
+
+    public final var syncOptions: NIOSynchronousChannelOptions? {
```
Lukasa, line 495: Same note about `final`.

Lukasa (changes requested):

Sources/NIOTransportServices/NIOTSConnectionChannel.swift
glbrntt: Oof, thanks, too much copy-pasta :(

Lukasa (approved):

Lukasa:
@swift-nio-bot test this please

Issue #92, opened 2020-06-09
NIOTS doesn't let you control the equivalent of `maxMessagesPerRead` and `recvAllocator`


weissi:
https://github.com/apple/swift-nio-transport-services/blob/2ac8fde712c1b1a147ecb7065824a40d2c09d0cb/Sources/NIOTransportServices/NIOTSConnectionChannel.swift#L583-L585

NIOTS always does one underlying `receive` per `read` and always up to `8192`.

So that kinda corresponds to:

- `maxMessagesPerRead = 1`
- `recvAllocator = AdaptiveRecvAllocator(minimum: 1, maximum: 8192, initial: ???)`

We should make those values configurable.

edwellbrook:
Hi @weissi! Just came across this in one of my own projects and have some motivation to fix it.

This would be my first time contributing to to an `swift-nio` project, though I have some basic understanding of how things work. Some guidance would be appreciated...

Could you suggest what a good user-facing API would look like for configuring these values? Perhaps another option under `NIOTSChannelOptions`?

Thanks!

weissi:
Hi @edwellbrook , awesome, and welcome! And thank you for your PR, I'll leave comments regarding your PR on the actual PR.

Were you separately from your existing PR (which configures the maximum read size) interested in working on `maxMessagesPerRead`?

edwellbrook:
Unfortunately I don't have much need for `maxMessagesPerRead` so I can't budget the time for it right now. If my existing PR needs updating I might be able to quickly fit this in though!

Not super familiar with `maxMessagesPerRead` so would need a suggestion on how it's supposed to work and should be implemented.

weissi:
@edwellbrook totally up to you. I fully understand if you can't budget the time. Ping me if you'd like an explanation of what it does exactly and how we could implement it.

edwellbrook:
@weissi It's looking like I may need `maxMessagesPerRead` after all (if I'm understanding it correctly).

Does it refers to the number of messages an `InboundChannelHandler` will receive from a `read` call before needing to call `read` again? i.e. setting `maxMessagesPerRead` to `1` would mean an `InboundChannelHandler` needs to call `read` for each message?

If that's right... I'd love a suggestion on how to implement it.

Cheers!

Lukasa:
Yup, that's exactly what it means.

[...]

PR #118, opened 2021-04-16
Set the max bytes to be read from a connection

edwellbrook:
Add a `NIOTSChannelOption` to configure the maximum number of bytes to be read from a connection.

### Motivation:

Currently the max number of bytes is hard coded to 1KB. In some scenarios API consumers may want to change this to better suit their project. See also: #92

### Modifications:

- Added a new struct to `NIOTSChannelOptions.Types` to represent the new option
- Added a new property to `ConnectionChannelOptions` to store the default or configured value
- Use the stored value when reading from the connection

### Result:

- Nothing will change to existing setups
- A new option will be available to API consumers to configure the max bytes to be read from a connection

### DIFF
--- a/Sources/NIOTransportServices/NIOTSChannelOptions.swift
+++ b/Sources/NIOTransportServices/NIOTSChannelOptions.swift
@@ -28,6 +28,8 @@ public struct NIOTSChannelOptions {
 
     public static let currentPath = NIOTSChannelOptions.Types.NIOTSCurrentPathOption()
 
+    public static let maxReceiveBytes = NIOTSChannelOptions.Types.NIOMaxReceiveBytesOption()
+
     public static let metadata = { (definition: NWProtocolDefinition) -> NIOTSChannelOptions.Types.NIOTSMetadataOption in
         .init(definition: definition)
     }
@@ -89,6 +91,13 @@ extension NIOTSChannelOptions {
         public struct NIOTSCurrentPathOption: ChannelOption, Equatable {
             public typealias Value = NWPath
             
+            public init() {}
+        }
+
+        @available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
+        public struct NIOMaxReceiveBytesOption: ChannelOption, Equatable {
+            public typealias Value = Int
+
             public init() {}
         }
         
--- a/Sources/NIOTransportServices/NIOTSConnectionChannel.swift
+++ b/Sources/NIOTransportServices/NIOTSConnectionChannel.swift
@@ -34,6 +34,9 @@ private struct ConnectionChannelOptions {
 
     /// Whether this channel should wait for the connection to become active.
     internal var waitForActivity: Bool = true
+
+    // Maximum number of bytes read from a connection
+    internal var maxRecvBytes: Int = 8192
 }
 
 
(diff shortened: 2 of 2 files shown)


Lukasa:
Thanks for this patch!

@weissi In your original issue you suggested using RecvByteBufferAllocator. I kinda think that would be the better choice here as it'd be more compatible with other NIO channels: what do you think?

Sources/NIOTransportServices/NIOTSChannelOptions.swift
```diff
@@ -28,6 +28,8 @@ public struct NIOTSChannelOptions {
 
     public static let currentPath = NIOTSChannelOptions.Types.NIOTSCurrentPathOption()
 
+    public static let maxReceiveBytes = NIOTSChannelOptions.Types.NIOMaxReceiveBytesOption()
```
weissi, line 31: @Lukasa I think a separate option (as suggested in this PR) is actually better than matching the socket channel with a RecvAllocator, wdyt? We can't sensibly control the strategy that NW.fw will use so I'm actually happy with this as is, wdyt?

weissi:
> Thanks for this patch!
> 
> @weissi In your original issue you suggested using RecvByteBufferAllocator. I kinda think that would be the better choice here as it'd be more compatible with other NIO channels: what do you think?

Oh sorry, I missed this. So yeah, initially I thought a `RecvByteBufferAllocator` would be better (because more in line with NIO). However, now I think I changed my mind because we may pointlessly over-allocate. Let's say the user installs the `AdaptiveRecvByteBufferAllocator`, that one chooses 64k, but then NW.fw only hands us say 1k. Then we just pointlessly over-allocated for no real reason (we'd copy just 1k into the 64k buffer). In NIO on sockets that makes sense because making better-fitting buffer would mean an extra alloc+copy but with NW.fw we need the alloc+copy anyway so I'd think picking the buffer that fits best makes more sense?

weissi:
@Lukasa If we could design `RecvByteBufferAllocator` againn, then we could make it have a method that returns the next size (instead of just `func buffer(allocator: ByteBufferAllocator) -> ByteBuffer`) and use that returned size for NW.fw.

What we _could_ obviously do is to add an optional requirement `func nextAllocationSize() -> Int?` and use that for NIOTS?

Lukasa:
Hmm. We kinda need a new allocator protocol for io_uring anyway, don't we? Or am I misunderstanding that?

weissi:
> Hmm. We kinda need a new allocator protocol for io_uring anyway, don't we? Or am I misunderstanding that?

I see, yes, that is a good point!

What about this PR then, would we want until we have a new allocator for io_uring (that would support not copying the `(Dispatch)Data` at all or should we move forward with either `.maxReceiveBytes` or adding a `.nextAllocSize()` to `RecvAllocator`?

Lukasa:
I think the new allocator protocol we need for io_uring is potentially very similar to the thing we want here. Might be good to address both tasks in one go? You could always _compose_ the old protocol from the new one, right, so pre-existing channels can kinda trivially support the new API.

I don't think adding a `nextAllocSize` is the way to go: I think a new type is probably better.

weissi:
> I think the new allocator protocol we need for io_uring is potentially very similar to the thing we want here.

Yes, agreed.


> Might be good to address both tasks in one go?

Fine by me. I think we should take the action to start designing that relatively soon though.


> I don't think adding a `nextAllocSize` is the way to go: I think a new type is probably better.

Ok, sounds good to me. Said that, an (optional) `nextAllocSize` doesn't look like too big of a wrinkle to me. But then, I do support your idea of designing a new allocator that supports this and io_uring in a really good way!

PR #127, opened 2021-08-19
Move check for event loop shutdown into taskQueue

PeterAdams-A:
Motivation:

The event loop state is not protected for access on different threads.
This means it must only be accessed from the task queue.

Modifications:

Move check for event loop shutdown into taskQueue

Result:

Event Loop state only accessed from the task queue.


PeterAdams-A:
Also see rdar://82043918
It feels slightly counter intuitive to check if the EventLoop is shutdown from on the loop but I believe this is fine as the dispatch queue doesn't go away with shutdown.  A similar pattern seems to be used elsewhere in this file too.

PeterAdams-A:
Raised night build fail - https://bugs.swift.org/browse/SR-15090

glbrntt (approved):

PR #144, opened 2022-05-04
Remove build_podspec.sh

glbrntt:
Motivation:

We no longer support Cocoapods

Modifications:

- Remove build_podspec.sh

Result:

Fewer unused files


Lukasa (approved):
Love a negative diff

PR #154, opened 2022-07-29
Use Docc for documentation

Lukasa:
Motivation

Documentation is nice, and we can help support users by providing useful
clear docs.

Modifications

Add Docc to 5.6 and later builds
Make sure symbol references work
Add overview docs

Result

Nice rendering docs

### DIFF
--- a/Sources/NIOTransportServices/NIOTSListenerBootstrap.swift
+++ b/Sources/NIOTransportServices/NIOTSListenerBootstrap.swift
@@ -17,7 +17,7 @@ import NIOCore
 import Dispatch
 import Network
 
-/// A `NIOTSListenerBootstrap` is an easy way to bootstrap a `NIOTSListenerChannel` when creating network servers.
+/// A ``NIOTSListenerBootstrap`` is an easy way to bootstrap a `NIOTSListenerChannel` when creating network servers.
 ///
 /// Example:
 ///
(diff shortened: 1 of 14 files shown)


Lukasa:
Uh, this is new. @yim-lee any idea what's happened to SwiftPM here? It seems to have generated a test file that emits a compiler warning, which we choke on because we use warnings as errors.

Sources/NIOTransportServices/NIOTSListenerBootstrap.swift
```diff
@@ -175,7 +175,7 @@ public final class NIOTSListenerBootstrap {
 
     /// Specifies a `ChannelOption` to be applied to the `NIOTSListenerChannel`.
     ///
-    /// - note: To specify options for the accepted `NIOTSConnectionChannels`s, look at `NIOTSListenerBootstrap.childChannelOption`.
+    /// > Note: To specify options for the accepted `NIOTSConnectionChannel`s, look at ``childChannelOption(_:value:)``.
```
FranzBusch, line 178: Just out of curiosity is `- Note` not working in Docc anymore?

FranzBusch (approved):
LGTM!

Lukasa:
Oh dang it, I know what's happening: this is because there _are no tests_ on Linux. Ok, I'll add a dummy test.

Sources/NIOTransportServices/NIOTSListenerBootstrap.swift
Lukasa: I dunno, they encourage use of this angle bracket version in the docs so I did that.

yim-lee:
🎉

PR #155, opened 2022-08-31
Adopt `Sendable`

dnadoba:
Bumps `swift-nio` to `2.41.1` to have access to `NIOPreconcurrencySendable`.

`NIOTSMetadataOption` should probably be marked non-Sendable because it contains a reference type `NWProtocolDefinition` (and the Value of type `NWProtocolMetadata` is also a reference type). Network.framework has not adopted Sendable so it is hard to say if it is Sendable or not.  However, `NIOTSMetadataOption` conforms to `ChannelOption` which inherits from `NIOPreconcurrencySendable` and therefore must be `Sendable`. We will need to decided what we do with that type but that can be done separately.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -21,7 +21,7 @@ let package = Package(
         .library(name: "NIOTransportServices", targets: ["NIOTransportServices"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/apple/swift-nio.git", from: "2.32.0"),
+        .package(url: "https://github.com/apple/swift-nio.git", from: "2.41.1"),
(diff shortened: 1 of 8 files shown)


Lukasa (approved):

Lukasa:
A few compiler errors need fixing up

dnadoba:
https://github.com/apple/swift-nio-transport-services/pull/155/commits/f0b67962c4853b44d3bdd6d94c43490254c4a7b9 workarounds a compiler bug. A variation of it is still presented in the latest compiler but doesn't effect us: https://github.com/apple/swift/issues/60876

PR #161, opened 2022-10-13
Remove `#if compiler(>=5.5)`

dnadoba:
### Motivation
We only support Swift 5.5.2+. 

### Modification
Remove all `#if swift(>=5.5)` conditional compilation blocks.

### Result
less branching


Sources/NIOTransportServices/NIOTSNetworkEvents.swift
```diff
@@ -103,7 +103,6 @@ public enum NIOTSNetworkEvents {
     }
 }
 
-#if swift(>=5.6)
 @available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
 extension NIOTSNetworkEvents.BetterPathAvailable: Sendable {}
 @available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
```
dnadoba, line 108: This says `5.6` but I think it should have been 5.5. Lets see if CI complains.

Lukasa (approved):

PR #164, opened 2022-10-28
Use #fileID/#filePath instead of #file

carolinacass:
Motivation:

#fileID introduced in Swift 5.3, so no longer need to use #file anywhere                                  

Modifications:

Changed #file to #filePath because of XCT Assertions

### DIFF
--- a/Tests/NIOTransportServicesTests/NIOTSEndToEndTests.swift
+++ b/Tests/NIOTransportServicesTests/NIOTSEndToEndTests.swift
@@ -21,7 +21,7 @@ import Network
 
 
 func assertNoThrowWithValue<T>(_ body: @autoclosure () throws -> T, defaultValue: T? = nil, message: String? = nil,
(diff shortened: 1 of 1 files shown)


Tests/NIOTransportServicesTests/NIOTSEndToEndTests.swift
```diff
@@ -21,7 +21,7 @@ import Network
 
 
 func assertNoThrowWithValue<T>(_ body: @autoclosure () throws -> T, defaultValue: T? = nil, message: String? = nil,
-                               file: StaticString = #file, line: UInt = #line) throws -> T {
+                               file: StaticString = #filePa, line: UInt = #line) throws -> T {
```
glbrntt, line 24: Oops 🙊

```suggestion
                               file: StaticString = #filePath, line: UInt = #line) throws -> T {
```

glbrntt (changes requested):

glbrntt (approved):

PR #165, opened 2022-11-08
Expose multicast service type

Lukasa:
Motivation

As we've rolled out support for multicast on Linux, it makes sense to add equivalent configuration for swift-nio-transport-services. The two features aren't one-to-one, as Network.framework has substantially more capability than the Linux functionality.

Modifications

- Expose a multipathServiceType channel option
- Add a test to confirm we use it properly

Result

Multicast service types are available.


glbrntt (approved):
LGTM

PR #180, opened 2023-07-13
NIOSingletonsTransportServices: Use NIOTS in easy mode

weissi:
### Motivation:

(companion PR to https://github.com/apple/swift-nio/pull/2471)

SwiftNIO allows and encourages to precisely manage all operating system resources like file descriptors & threads. That is a very important property of the system but in many places -- especially single Swift Concurrency arrived -- many simpler SwiftNIO programs only require a single, globally shared EventLoopGroup. Often even with just one thread.

Long story short: Many, probably most users would happily trade precise control over the threads for not having to pass around `EventLoopGroup`s. Today, many of those users resort to creating (and often leaking) threads because it's simpler. Adding singletons type which _lazily_ provide a singleton `EventLoopGroup` and an `NIOThreadPool` is IMHO a much better answer.

### Modifications:

- Add `NIOTSEventLoopGroup.singleton`
- Add `NIOSingletons.transportServicesEventLoopGroup`

### Result:

- Easier use of NIO that requires fewer parameters for users who don't require full control.
- Helps with https://github.com/apple/swift-nio/issues/2142

### DIFF
--- a/Sources/NIOTransportServices/NIOTSSingletons.swift
+++ b/Sources/NIOTransportServices/NIOTSSingletons.swift
@@ -1,0 +1,88 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2023 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#if canImport(Network)
+import NIOCore
+
+@available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
+extension NIOTSEventLoopGroup {
+    /// A globally shared, lazily initialized, singleton ``NIOTSEventLoopGroup``.
+    ///
+    /// SwiftNIO allows and encourages the precise management of all operating system resources such as threads and file descriptors.
+    /// Certain resources (such as the main `EventLoopGroup`) however are usually globally shared across the program. This means
+    /// that many programs have to carry around an `EventLoopGroup` despite the fact they don't require the ability to fully return
+    /// all the operating resources which would imply shutting down the `EventLoopGroup`. This type is the global handle for singleton
+    /// resources that applications (and some libraries) can use to obtain never-shut-down singleton resources.
+    ///
+    /// Programs and libraries that do not use these singletons will not incur extra resource usage, these resources are lazily initialized on
+    /// first use.
+    ///
+    /// The loop count of this group is determined by `NIOSingletons.groupLoopCountSuggestion`.
+    ///
+    /// - note: Users who do not want any code to spawn global singleton resources may set
+    ///         `NIOSingletons.singletonsEnabledSuggestion` to `false` which will lead to a forced crash
+    ///         if any code attempts to use the global singletons.
+    public static var singleton: NIOTSEventLoopGroup {
+        return NIOSingletons.transportServicesEventLoopGroup
+    }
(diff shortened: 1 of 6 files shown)


weissi:
This cannot be merged until https://github.com/apple/swift-nio/pull/2471 has been released in NIO 2.57.0

Lukasa (changes requested):
Just putting a hold on this for now.

Sources/NIOTransportServices/NIOTSSingletons.swift
```diff
@@ -0,0 +1,79 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#if canImport(Network)
+import NIOCore
+
+@available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
+extension NIOTSEventLoopGroup {
```
glbrntt, line 19: Could you add an extension to `EventLoopGroup` like https://github.com/weissi/swift-nio/blob/2daebdcff77abba902cfdc33eac4e058fe8f3ae8/Sources/NIOPosix/PosixSingletons.swift#L57-L64?

glbrntt (commented):
Looks good aside from one nit and the missing NIO release

Sources/NIOTransportServices/NIOTSSingletons.swift
weissi: done!

FranzBusch (approved):

weissi:
With the NIO 2.58.0 release this should now go green

FranzBusch (approved):

weissi:
@Lukasa merging is currently blocked because it still has you down as a 'request changes', at your convenience PTAL

glbrntt (approved):

Sources/NIOTransportServices/NIOTSSingletons.swift
```diff
@@ -0,0 +1,88 @@
...
+    /// - note: Users who do not want any code to spawn global singleton resources may set
+    ///         `NIOSingletons.singletonsEnabledSuggestion` to `false` which will lead to a forced crash
+    ///         if any code attempts to use the global singletons.
+    public static var singleton: NIOTSEventLoopGroup {
+        return NIOSingletons.transportServicesEventLoopGroup
+    }
+}
+
+@available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
+extension EventLoopGroup where Self == NIOTSEventLoopGroup {
```
Lukasa, line 42: Does this extension ever exist where the other does not?

weissi: @glbrntt suggested this one. The answer is yes, in the many places where we take `any EventLoopGroup`. there you can now type `.singletonMultiThreadedEventLoopGroup`/`.singletonTransportServicesEventLoopGroup`.

I don't know if I love it but NIO released this pattern and even if we decided that it's bad, it's an easy one to deprecated. Given that it's in NIO we should also have it in TS.

glbrntt: FWIW I think this is the more useful API for most users because it's easy to discover: IDEs will suggest this when you dot-tab-complete for `any EventLoopGroup`. If you didn't have this API users would have to know about the ELG types or singleton objects to find them.

Lukasa: That's deeply weird.

Lukasa (approved):

Sources/NIOTransportServices/NIOTSSingletons.swift
glbrntt: I think we have SwiftUI to thank for this 😃 e.g. https://developer.apple.com/documentation/swiftui/view/labelstyle(_:)

PR #181, opened 2023-07-26
Adopt latest SPI(AsyncChannel) changes

FranzBusch:
# Motivation
We introduced some breaking SPI(AsyncChannel) changes in NIO that we have to adopt here.

# Modification
This PR adopts the latest `NIOProtocolNegotiationResult` APIs. Additionally, it also drops all bind/connect methods on the bootstraps that are specific to protocol negotiation or `NIOAsyncChannel`.

# Result
Green CI on `main` and alignment between `NIOPosix` and `NIOTS.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -21,7 +21,7 @@ let package = Package(
         .library(name: "NIOTransportServices", targets: ["NIOTransportServices"]),
     ],
     dependencies: [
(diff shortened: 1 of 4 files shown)


dnadoba (approved):
Nice cleanup!

Issue #133, opened 2021-10-07
copy-paste error in NIOTSConnectionChannel getOption0 allowLocalEndpointReuse


vvuk:
Looks like a copy-paste mistake here:

https://github.com/apple/swift-nio-transport-services/blob/e7f5278a26442dc46783ba7e063643d524e414a0/Sources/NIOTransportServices/NIOTSConnectionChannel.swift#L320-L323

`allowLocalEndpointReuse` is being set to the value of `NIOTSEnablePeerToPeerOption`, not itself.

Though it might not really matter, because from everything I can tell `allowLocalEndpointReuse` seems to be completely broken on NWListener (ref https://developer.apple.com/forums/thread/129452).

AgentFeeble:
It looks like a similar issue slipped through in

AgentFeeble:
It looks like a similar issue slipped through in `StateManagedNWConnectionChannel`

https://github.com/apple/swift-nio-transport-services/blob/251e9e76134570f0cc5c21e5c6ff3ccb247472ab/Sources/NIOTransportServices/StateManagedNWConnectionChannel.swift#L540C9-L541C113

    case is NIOTSChannelOptions.Types.NIOTSAllowLocalEndpointReuse:
        self.allowLocalEndpointReuse = value as! NIOTSChannelOptions.Types.NIOTSEnablePeerToPeerOption.Value

It won't make any difference as long as both `NIOTSEnablePeerToPeerOption.Value` and `NIOTSAllowLocalEndpointReuse.Value` are `Bool`s, but it's probably better to fix it anyway for consistency.

Lukasa:
Agreed. Do you feel like making a patch?

PR #186, opened 2023-09-22
Fix a typo when casting the allowLocalEndpointReuse channel option value

AgentFeeble:
This PR addresses issue #133.

### Motivation:

There was a typo when setting the allowLocalEndpointReuse channel option, where it was cast to NIOTSEnablePeerToPeerOption.Value instead of NIOTSAllowLocalEndpointReuse.Value. Both of these types are Bools, so this wouldn't cause any actual issues, this change is to more to keep the code consistent, and reduce confusion for future contributors or others reading the code base.

### Modifications:

This commit casts the allowLocalEndpointReuse property in StateManagedListenerChannel and StateManagedNWConnectionChannel.

### Result:

The typo is fixed after this change.


Lukasa (approved):
Nice one, thanks!

Lukasa:

PR #187, opened 2023-09-28
Fire the pipelines  error caught method when NWConnection's state changes to failed

Cartisim:
Fire the pipelines  error caught method when NWConnection's state changes to failed

There seems to need to be a method fired when NSConnection's state changes to failed in order to notify other handlers in the pipeline.

### Motivation:

When NWConnection fails the connection due to a number of reasons, one being that the clients devices loses internet and we want to receive the failed error in order to handle the applications behavior when the internet is lost. We need to fire the `errorCaught` method in order to notify handler's in the pipeline.

### Modifications:

On line 277 of the file StateManagedNWConnectionChannel.swift I added the line of code to forward the error, namely...
` self.pipeline.fireErrorCaught(error)`

### Result:

Our Handlers can now receive the error and handle the socket error state accordingly.


Cartisim:
Done ✅, Hopefully the test is sufficient. Thanks a bunch!

Cartisim:
@Lukasa I am happy to do what I can to push this one through. Thanks for your work.

Sources/NIOTransportServices/StateManagedNWConnectionChannel.swift
```diff
@@ -259,19 +259,25 @@ extension StateManagedNWConnectionChannel {
...
+        
         // Step 1 is to tell the network stack we're done.
         // TODO: Does this drop the connection fully, or can we keep receiving data? Must investigate.
         conn.cancel()
-
+        
         // Step 2 is to fail all outstanding writes.
         self.dropOutstandingWrites(error: error)
-
+
```
Lukasa, line 269: Can you remove the added whitespace here?

Cartisim: It appears that on the main branch lines 262, 266, and 269 all have a single whitespace for those lines. Did I misunderstand what was needed?

Lukasa: You added some spaces. :smile:

Tests/NIOTransportServicesTests/NIOTSConnectionChannelTests.swift
```diff
@@ -865,5 +865,61 @@ class NIOTSConnectionChannelTests: XCTestCase {
...
+                XCTAssertNotNil(error)
+                testCompletePromise.succeed(error!)
+            }
+        }
+        
+        let listener = try NIOTSListenerBootstrap(group: self.group)
+            .childChannelInitializer { channel in
+                return channel.eventLoop.makeSucceededVoidFuture()
+            }
+            .bind(host: "127.0.0.1", port: 8080)
```
Lukasa, line 906: Can we `bind` `0` instead? This can cause awkward conflicts if this port is in use somewhere else.

Tests/NIOTransportServicesTests/NIOTSConnectionChannelTests.swift
```diff
@@ -865,5 +865,61 @@ class NIOTSConnectionChannelTests: XCTestCase {
...
+            }
+            func channelActive(context: ChannelHandlerContext) {
+                let listenerErrorPromise = self.listenerChannel.eventLoop.next().makePromise(of: Error.self)
+                listenerChannel
+                    .close()
+                    .whenSuccess { _ in
+                        context.channel.write(ByteBuffer(data: Data()))
+                            .whenFailure({ error in
+                                listenerErrorPromise.succeed(error)
+                            })
```
Lukasa, line 889: I don't think this path should complete the promise: we're testing for this happening in errorcaught, not on the write promise.

Cartisim: Are you cool with me just throwing the result of the write away?

Lukasa (approved):
Nice one, thanks!

Lukasa:
Can you bring your branch up-to-date with main?

Cartisim:
@Lukasa  ✅ 
Thanks again for your hard work!

Lukasa (approved):

Lukasa:

PR #190, opened 2023-10-09
Implement customization point for `enqueue`

FranzBusch:
# Motivation
The default implementation of `SerialExecutor` for `EventLoop`s causes a lot of allocations that's why we added a new customization point to allow `EventLoop`s to implement fast paths.

# Modification
This PR adds the customization point to `NIOTSEventLoop`.

### DIFF
--- a/Sources/NIOTransportServices/NIOTSEventLoop.swift
+++ b/Sources/NIOTransportServices/NIOTSEventLoop.swift
@@ -262,4 +262,21 @@ extension NIOTSEventLoop {
         assert(oldChannel != nil)
     }
 }
+
+// MARK: SerialExecutor conformance
+#if compiler(>=5.9)
+@available(macOS 14.0, *)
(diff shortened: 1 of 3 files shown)


Sources/NIOTransportServices/NIOTSEventLoop.swift
```diff
@@ -262,4 +262,21 @@ extension NIOTSEventLoop {
...
     }
 }
+
+// MARK: SerialExecutor conformance
+#if compiler(>=5.9)
+@available(macOS 14.0, *)
+extension NIOTSEventLoop: NIOSerialEventLoopExecutor {
+    func enqueue(_ job: consuming ExecutorJob) {
+        let unownedJob = UnownedJob(job)
+        if let executorQueue = self.taskQueue as? _DispatchSerialExecutorQueue {
```
FranzBusch, line 272: This is currently just available as underscored API and cannot be relied on. This PR is more for demonstrating that this can be fast pathed as well.

PR #202, opened 2024-05-13
Fix the syncOptions on most channels

Lukasa:
Motivation:

Looks like when we previously added syncOptions support to our channels, we had a few issues. The listeners had code added, but the code never worked. This is because the code was defined in subclasses, but the protocol conformance comes from the parent class, and that parent class did not have a customized protocol witness.

The datagram channel was also entirely missing its support.

Modifications:

- Added the missing unit tests for sync options.
- Added syncOptions to StateManagedListenerChannel base class.
- Added overrides to the listener subclasses.
- Added syncOptions to datagram channel

Result:

Sync options actually work across the board.


FranzBusch (approved):
LGTM!

PR #203, opened 2024-05-13
Add ChannelOptions to extract base types.

Lukasa:
Motivation:

In some cases users may want to access APIs we haven't exposed in NIOTransportServices. We should have a fallback that allows users to do this.

Modifications:

- Add ChannelOptions for getting NWConnection and NWListener.

Result:

Users have an escape hatch


FranzBusch (approved):
LGTM!

PR #205, opened 2024-08-22
Add configuration of multipathServiceType in NIOTSConnectionBootstrap

Aperence:
### Motivation:

Allow different devices to leverage the capabilities of Mutipath TCP (MPTCP) to enhance the network reliability. Thanks to MPTCP, a connection can for example automatically migrate to another interface if it deteriorates on the first one. This can be especially interesting on MacOS X and iOS, where devices may frequently benefit from multiple interfaces (ethernet + Wi-Fi for Macs and Wi-fi + cellular for iOS). Allowing developers to enable MPTCP on their connections seems thus like a fine addition to this library.

### Modifications:

Added a function "withMultipath" on NIOTSConnectionBootstrap, that allow to configure the type of service used for MPTCP (defaults to disabled). This value will be stored in a field, and then propagated to the underlying channel when the connect method will be called. Also updated the parameters field of NIOTSConnectionChannel to set the multipathServiceType accordingly.

### Result:

Users will now be able to easily enable Multipath TCP, allowing them to benefit from seamless handover, interactive mode to automatically use the lowest delay interface or aggregate mode to send data in parallel on both interfaces.

Example: 
```swift
let group = NIOTSEventLoopGroup()
defer {
     try! group.syncShutdownGracefully()
}
let bootstrap = NIOTSConnectionBootstrap(group: group)
    .withMultipath(.handover)   // set handover mode
    .channelInitializer { channel in
        channel.pipeline.addHandler(MyChannelHandler())
    }
try! bootstrap.connect(host: "example.org", port: 12345).wait()
/* the Channel is now connected */
```

### DIFF
--- a/Sources/NIOTransportServices/NIOTSConnectionBootstrap.swift
+++ b/Sources/NIOTransportServices/NIOTSConnectionBootstrap.swift
@@ -154,6 +154,12 @@ public final class NIOTSConnectionBootstrap {
         return self
     }
 
+    /// Specifies a type of Multipath service to use for this connection, instead of the default
+    /// service type for the event loop.
+    public func withMultipath(_ type: NWParameters.MultipathServiceType) -> Self {
+        self.channelOption(NIOTSChannelOptions.multipathServiceType, value: type)
+    }
+
     /// Specify the `host` and `port` to connect to for the TCP `Channel` that will be established.
     ///
     /// - parameters:
--- a/Tests/NIOTransportServicesTests/NIOTSBootstrapTests.swift
+++ b/Tests/NIOTransportServicesTests/NIOTSBootstrapTests.swift
@@ -336,6 +336,25 @@ final class NIOTSBootstrapTests: XCTestCase {
             try? connectionChannel?.close().wait()
         }
     }
+
+    func testBootstrapsMultipath() throws {
+        let group = NIOTSEventLoopGroup()
+        defer {
+            try! group.syncShutdownGracefully()
+        }
+
+        let listenerBootstrap = NIOTSListenerBootstrap(group: group).withMultipath(.handover)
+        let connectionBootstrap = NIOTSConnectionBootstrap(group: group).withMultipath(.handover)
+
+        let listenerChannel: Channel = try listenerBootstrap.bind(host: "localhost", port: 0).wait()
+        let connectionChannel: Channel = try connectionBootstrap.connect(to: listenerChannel.localAddress!).wait()
+        defer{
+            try? listenerChannel.close().wait()
+            try? connectionChannel.close().wait()
+        }
+        XCTAssertEqual(try listenerChannel.getOption(NIOTSChannelOptions.multipathServiceType).wait(), .handover)
+        XCTAssertEqual(try connectionChannel.getOption(NIOTSChannelOptions.multipathServiceType).wait(), .handover)
+    }
 }
 
 extension Channel {
--- a/Sources/NIOTransportServices/NIOTSListenerBootstrap.swift
+++ b/Sources/NIOTransportServices/NIOTSListenerBootstrap.swift
@@ -235,6 +235,16 @@ public final class NIOTSListenerBootstrap {
         return self
     }
 
+    /// Specifies a type of Multipath service to use for this listener, instead of the default
+    /// service type for the event loop.
+    ///
+    /// Warning: Multipath service doesn't seem supported on the listener side yet, as
+    /// described on https://www.mptcp.dev/macOS.html. Note that enabling Multipath
+    /// may then generate unexpected errors, use this function with caution.
+    public func withMultipath(_ type: NWParameters.MultipathServiceType) -> Self {
+        self.serverChannelOption(NIOTSChannelOptions.multipathServiceType, value: type)
+    }
+
     /// Bind the `NIOTSListenerChannel` to `host` and `port`.
     ///
     /// - parameters:


Lukasa (changes requested):
Thanks for filing this! This is a nice addition, as offering a convenience API for multipath service type is useful. However, the multipath service type is _already_ configurable via the `NIOTSChannelOptions.multipathServiceType` option. I'm all for offering a convenience abstraction for this, but I recommend we have it decay down to just setting the channel option, as that will ensure we only have a single path for getting this to work correctly.

Aperence:
Thank you for your feedback, I'll try to simplify it to use the channel options !

Aperence:
Hello, 

I've simplified (a lot) the code thanks to your feedback, it now uses the `IOTSChannelOptions`, limiting the drastically the modifications needed

Sources/NIOTransportServices/NIOTSConnectionChannel.swift
```diff
@@ -157,7 +157,9 @@ internal final class NIOTSConnectionChannel: StateManagedNWConnectionChannel {
     internal var connectPromise: EventLoopPromise<Void>?
 
     internal var parameters: NWParameters {
-        NWParameters(tls: self.tlsOptions, tcp: self.tcpOptions)
+        let parameters = NWParameters(tls: self.tlsOptions, tcp: self.tcpOptions)
+        parameters.multipathServiceType = self.multipathServiceType
+        return parameters
```
Lukasa, line 162: Is this change actually needed? I'd have thought it would work fine without it.

Sources/NIOTransportServices/NIOTSConnectionBootstrap.swift
```diff
@@ -154,6 +154,12 @@ public final class NIOTSConnectionBootstrap {
         return self
     }
 
+    /// Specifies a type of Multipath service to use for this connection, instead of the default 
+    /// service type for the event loop.
+    public func withMultipath(_ type: NWParameters.MultipathServiceType) -> Self {
```
Lukasa, line 159: We should probably add this to the listener bootstrap too.

Sources/NIOTransportServices/NIOTSConnectionChannel.swift
Aperence: I've inspected a bit more the code, and it seems that these changes aren't needed indeed

Sources/NIOTransportServices/NIOTSConnectionBootstrap.swift
Aperence: I've added this function to listeners too. Note however that [MPTCP doesn't seem to be supported](https://www.mptcp.dev/macOS.html) on the server side for the moment (as it generates warnings when trying to connect to this socket). Perhaps adding a warning about it somewhere would help ?

Lukasa: Yeah, that's a good idea. Let's add a warning to the doc comment on the listener bootstrap.

Aperence: Done in the latest commit ! Thank you for your feedback !

Lukasa (commented):
Thanks for this, and sorry for the delay in getting to it! Can you briefly add a unit test that configures both a listener and a connection bootstrap and connects them, and checks the set values using the channel options? I'd just like to make sure we don't accidentally regress this API in future.

Aperence:
No problem, I've just added this test now !

Tests/NIOTransportServicesTests/NIOTSBootstrapTests.swift
```diff
@@ -336,6 +336,29 @@ final class NIOTSBootstrapTests: XCTestCase {
...
+        let group = NIOTSEventLoopGroup()
+        defer {
+            try! group.syncShutdownGracefully()
+        }
+
+        let listenerBootstrap = NIOTSListenerBootstrap(group: group).withMultipath(.handover)
+        let connectionBootstrap = NIOTSConnectionBootstrap(group: group).withMultipath(.handover)
+
+        do{
+            var listenerChannel: Channel = try listenerBootstrap.bind(host: "localhost", port: 8000).wait()
```
Lukasa, line 350: Let's not force port 8000: instead, let's bind port 0 and then use `localAddress` to get the target for the connection.

Aperence: It will be better indeed !

Tests/NIOTransportServicesTests/NIOTSBootstrapTests.swift
```diff
@@ -336,6 +336,29 @@ final class NIOTSBootstrapTests: XCTestCase {
...
+            let connectionChannel: Channel = try connectionBootstrap.connect(to: listenerChannel.localAddress!).wait()
+            defer{
+                try? listenerChannel.close().wait()
+                try? connectionChannel.close().wait()
+            }
+            XCTAssertEqual(try listenerChannel.getOption(NIOTSChannelOptions.multipathServiceType).wait(), .handover)
+            XCTAssertEqual(try connectionChannel.getOption(NIOTSChannelOptions.multipathServiceType).wait(), .handover)
+        }catch{
+            XCTFail()
+        }
```
Lukasa, line 360: We don't need this do-catch, the test will fail if it throws an error.

Aperence: Indeed !

Lukasa (approved):

Lukasa:

Lukasa:
Thank you very much @Aperence, and thanks for being very game with the feedback!

PR #207, opened 2024-09-09
Added support for connection viability updates

Cartisim:
# Enhancement: Integration of Viability Handler in `Network.framework`

We have successfully integrated the `viabilityUpdateHandler` from `Network.framework` into `NIOTransportServices`, enabling developers to leverage the full potential of the `Network.framework` API.

This enhancement provides adopters with critical insights into the viability of connections to remote endpoints. By being informed about the current network status, developers can implement a more responsive and adaptive approach to network behavior and conditions.

## Modifications

- Introduced two new methods along with corresponding calls to these methods within the `Network.framework` handler.
- These handlers trigger the `NIO` *InboundEventsTriggered* method, allowing consumers to respond to network changes effectively.
- Developed two structs within *NIOTSNetworkEvents* to accompany the *InboundEventsTriggered* methods, ensuring seamless data transfer.
- Conformed these structs to Sendability for compatibility with the required operating systems.

## Result

Clients can now receive viability events from `Network.framework`, enhancing their ability to manage network connections dynamically

### DIFF
--- a/Sources/NIOTransportServices/NIOTSNetworkEvents.swift
+++ b/Sources/NIOTransportServices/NIOTSNetworkEvents.swift
@@ -52,7 +52,7 @@ public enum NIOTSNetworkEvents {
         public var isViable: Bool
         
         /// Create a new ``NIOTSNetworkEvents/ViabilityUpdate`` event.
-        public init(isViable: Bool){
+        public init(isViable: Bool) {
             self.isViable = isViable
         }
     }
--- a/Tests/NIOTransportServicesTests/NIOTSEndToEndTests.swift
+++ b/Tests/NIOTransportServicesTests/NIOTSEndToEndTests.swift
@@ -563,11 +563,9 @@ class NIOTSEndToEndTests: XCTestCase {
             typealias InboundIn = ByteBuffer
             
             private let testCompletePromise: EventLoopPromise<Bool>
-            let listenerChannel: Channel
             
-            init(testCompletePromise: EventLoopPromise<Bool>, listenerChannel: Channel) {
+            init(testCompletePromise: EventLoopPromise<Bool>) {
                 self.testCompletePromise = testCompletePromise
-                self.listenerChannel = listenerChannel
             }
 
             func userInboundEventTriggered(context: ChannelHandlerContext, event: Any) {
@@ -589,15 +587,20 @@ class NIOTSEndToEndTests: XCTestCase {
             .channelInitializer { channel in
                 channel.pipeline.addHandler(
                     ViabilityHandler(
-                        testCompletePromise: testCompletePromise,
-                        listenerChannel: listener
+                        testCompletePromise: testCompletePromise
                     )
                 )
             }
             .connect(to: listener.localAddress!)
             .wait()
+        
+        do {
+            let result = try testCompletePromise.futureResult.wait()
+            XCTAssertEqual(result, true)
+        } catch {
+            XCTFail("Threw unexpected error \(error)")
+        }
         XCTAssertNoThrow(try connection.close().wait())
-        XCTAssertNoThrow(try testCompletePromise.futureResult.wait())
         XCTAssertNoThrow(try listener.close().wait())
     }
 }


Sources/NIOTransportServices/NIOTSNetworkEvents.swift
```diff
@@ -43,6 +43,20 @@ public enum NIOTSNetworkEvents {
         /// Create a new ``NIOTSNetworkEvents/BetterPathUnavailable`` event.
         public init(){ }
     }
+    
+    /// ``IsViable`` is triggered when the OS informs NIO that communication
+    /// with the remote endpoint is possible, indicating that the connection is viable.
+    public struct IsViable: NIOTSNetworkEvent {
```
glbrntt, line 49: I know this follows the pattern for `BetterPathAvailable`/`BetterPathUnavailable`, but is there any reason to not have a single event struct, e.g. `ViabilityUpdate` with an `isViable` flag?

glbrntt (commented):
Thanks for doing this, it seems like a useful addition.

Presumably we get a notification when a connection is established? If so it means we should be able to write a test that verifies we actually get the notification. Do you think you'd be able to add one? It looks like [`NIOTSEndToEndTests.swift`](https://github.com/apple/swift-nio-transport-services/blob/main/Tests/NIOTransportServicesTests/NIOTSEndToEndTests.swift) would be a good place to start.

Sources/NIOTransportServices/NIOTSNetworkEvents.swift
Cartisim: I agree I can make the name change, I just wanted to follow the current the pattern. There are a handful of things I would like to update with NIOTS, but just picking a few things off at a time.  I also didn't write any tests because I didn't see any others for the other handlers, but maybe I missed them. Give me an hour in the morning and I will have an updated PR.

glbrntt: Maybe @Lukasa has some insight into why there are separate events types for path availability?

Cartisim: I kind of like the Idea of using an enum. What do you think?

```swift
private func viabilityUpdateHandler(viable: Bool) {
        self.pipeline.fireUserInboundEventTriggered(NIOTSNetworkEvents.ViablitiyUpdate(viable: viable ? .isViable : .isNotViable))
    }

    /// ``IsViable`` is triggered when the OS informs NIO that communication
    /// with the remote endpoint is possible, indicating that the connection is viable.
    public struct ViablitiyUpdates: NIOTSNetworkEvent {
        
        public let viability: Viability
        
        public enum Viablitiy {
            case .isViable, .isNotViable
        }
        
        /// Create a new ``NIOTSNetworkEvents/IsViable`` event.
        public init(viability: Viability){
            self.viability = viability
        }
    }
```

From my viewpoint this is cleaner and nicer to work with. Would this be an acceptable approach for NIOTransportServices? @Lukasa @glbrntt

glbrntt: This is just a `Bool` with a fancy hat on 🤠 – I'd prefer to keep things simple here and stick with an `isViable` `Bool`.

Cartisim: Good point. I think this is a bit better now.

Tests/NIOTransportServicesTests/NIOTSEndToEndTests.swift
```diff
@@ -557,5 +557,49 @@ class NIOTSEndToEndTests: XCTestCase {
...
+            let listenerChannel: Channel
+            
+            init(testCompletePromise: EventLoopPromise<Bool>, listenerChannel: Channel) {
+                self.testCompletePromise = testCompletePromise
+                self.listenerChannel = listenerChannel
+            }
+
+            func userInboundEventTriggered(context: ChannelHandlerContext, event: Any) {
+                if event is NIOTSNetworkEvents.ViabilityUpdate {
+                    let update = event as! NIOTSNetworkEvents.ViabilityUpdate
```
glbrntt, line 575: You can just use `if let as` to avoid the `as!`:

```suggestion
                if let update = event as? NIOTSNetworkEvents.ViabilityUpdate {
```

Tests/NIOTransportServicesTests/NIOTSEndToEndTests.swift
```diff
@@ -557,5 +557,49 @@ class NIOTSEndToEndTests: XCTestCase {
                 print(error)
         }
     }
+    
+    func testViabilityUpdate() throws {
+        final class ViabilityHandler: ChannelInboundHandler {
+            typealias InboundIn = ByteBuffer
+            
+            private let testCompletePromise: EventLoopPromise<Bool>
+            let listenerChannel: Channel
```
glbrntt, line 566: This one isn't used

Sources/NIOTransportServices/NIOTSNetworkEvents.swift
```diff
@@ -43,6 +43,19 @@ public enum NIOTSNetworkEvents {
         /// Create a new ``NIOTSNetworkEvents/BetterPathUnavailable`` event.
         public init(){ }
     }
+    
+    /// ``ViabilityUpdate`` is triggered when the OS informs NIO that communication
+    /// with the remote endpoint is possible, indicating that the connection is viable.
+    public struct ViabilityUpdate: NIOTSNetworkEvent {
+
+        /// The current viability for the connection
+        public let isViable: Bool
```
Lukasa, line 52: Minor nit, this should probably be var.

```suggestion
        public var isViable: Bool
```

Sources/NIOTransportServices/NIOTSNetworkEvents.swift
```diff
@@ -106,6 +119,8 @@ extension NIOTSNetworkEvents.BetterPathAvailable: Sendable {}
 @available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
 extension NIOTSNetworkEvents.BetterPathUnavailable: Sendable {}
 @available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
+extension NIOTSNetworkEvents.ViabilityUpdate: Sendable {}
```
Lukasa, line 122: This extension is always `Sendable`, so it doesn't need the `@available`.

Tests/NIOTransportServicesTests/NIOTSEndToEndTests.swift
```diff
@@ -557,5 +557,49 @@ class NIOTSEndToEndTests: XCTestCase {
...
+                    ViabilityHandler(
+                        testCompletePromise: testCompletePromise,
+                        listenerChannel: listener
+                    )
+                )
+            }
+            .connect(to: listener.localAddress!)
+            .wait()
+        XCTAssertNoThrow(try connection.close().wait())
+        XCTAssertNoThrow(try testCompletePromise.futureResult.wait())
```
glbrntt, line 601: When this completes is `isViable` `true` or `false`? Can we assert what it is in the test? 

Assuming it becomes `true` when the connection is established can we wait for that to happen before closing the connection.

glbrntt (changes requested):
Left a couple of small comments but this looks great otherwise

Sources/NIOTransportServices/NIOTSNetworkEvents.swift
Cartisim: Isn't NIOTSNetworkEvents limited to `@available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *)`

Cartisim: Any suggestions? @Lukasa

glbrntt: @Cartisim you are correct here, the availability is required because of `NIOTSNetworkEvents`

glbrntt (approved):
This looks good to me, thank you! @Lukasa did you want to look as well?

Lukasa (approved):
Nice one, thanks @Cartisim!

PR #209, opened 2024-10-08
Don't depend on NIOFoundationCompat in NIOTransportServices on Linux

Cyberbeni:
Don't depend on NIOFoundationCompat in NIOTransportServices on Linux

### Motivation:

I'm trying to build a small utility tool based on mqtt-nio which depends on NIOTransportServices. I noticed that NIOTransportServices depends on NIOFoundationCompat on all platforms but only imports it where the Network framework is available. Removing this dependency on non-Apple platforms allows us to not import Foundations, significantly reducing the size of the build product when using the Swift Static Linux SDK.

### Modifications:

This PR and a similar one to mqtt-nio.

### Result:

Hello world docker image depending on mqtt-nio went from 169MB to 85MB.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -15,6 +15,12 @@
 
 import PackageDescription
 
+#if compiler(>=5.9)
+let applePlatforms: [Platform] = [.iOS, .macOS, .tvOS, .watchOS, .macCatalyst, .driverKit, .visionOS]
+#else
+let applePlatforms: [Platform] = [.iOS, .macOS, .tvOS, .watchOS, .macCatalyst, .driverKit]
+#endif
+
 let package = Package(
     name: "swift-nio-transport-services",
     products: [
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -30,7 +30,7 @@ let package = Package(
             dependencies: [
                 .product(name: "NIO", package: "swift-nio"),
                 .product(name: "NIOCore", package: "swift-nio"),
-                .product(name: "NIOFoundationCompat", package: "swift-nio"),
+                .product(name: "NIOFoundationCompat", package: "swift-nio", condition: .when(platforms: [.iOS, .macOS, .tvOS, .watchOS, .macCatalyst, .driverKit])),
```
FranzBusch, line 33: We need `visionOS` here as well. That is introduced in a later tools version so we might need multiple `Package.swift` files here.

FranzBusch:
Thanks for opening this PR. Conditionalising the dependency on an import makes sense!

Package.swift
Cyberbeni: Adding `#if compiler` check seems to work based on running `docker run --rm --volume .:/workspace swift:5.9.0 swift package resolve --package-path /workspace` and same with `swift:5.8`

FranzBusch:

FranzBusch:
@swift-server-bot add to allowlist

FranzBusch:

Lukasa:

FranzBusch (approved):

PR #210, opened 2024-10-14
Revert "Don't depend on NIOFoundationCompat in NIOTransportServices on Linux (#209)"

Lukasa:
This reverts commit 40ffcdef465d0870113b531b08454fe1712c01de in PR #209.

Unfortunately, Swift's behaviour around imports tends to be somewhat "leaky". In this case, the leak is that the dependency on the Package.swift allows downstream projects to import `NIOFoundationCompat` without needing to actually specify the package dependency. This has affected Hummingbird, which we noticed on our internal integration testing functionality:

```
hummingbird/Sources/Hummingbird/Codable/JSON/JSONCoding.swift:18:8: error: no such module 'NIOFoundationCompat'
import NIOFoundationCompat
       ^
```

cc @Joannis @adam-fowler for the Hummingbird report.

Unfortunately, we can't make this change until we're willing to use a semver major to achieve it. There may be some argument for doing that now, as the semver major will be very cheap to adopt across the ecosystem. But we'll need to do this in a considered way.

@Cyberbeni please feel free to reopen your PR targetting main, where we can discuss whether this is worth issuing a semver major for.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -15,12 +15,6 @@
 
 import PackageDescription
 
-#if compiler(>=5.9)
-let applePlatforms: [Platform] = [.iOS, .macOS, .tvOS, .watchOS, .macCatalyst, .driverKit, .visionOS]
-#else
-let applePlatforms: [Platform] = [.iOS, .macOS, .tvOS, .watchOS, .macCatalyst, .driverKit]
-#endif
-
 let package = Package(
     name: "swift-nio-transport-services",
     products: [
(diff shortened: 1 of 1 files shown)


FranzBusch (approved):

adam-fowler:
Ah that's frustrating as those imports are not needed. They were required by code that got removed. I've removed them from HB https://github.com/hummingbird-project/hummingbird/pull/587

adam-fowler:
Would a better solution be to make `NIOFoundationCompat` use `FoundationEssentials` if it is available, or is that also going to have leak issues.

Lukasa:
> Would a better solution be to make NIOFoundationCompat use FoundationEssentials if it is available, or is that also going to have leak issues.

Sadly, that also has leak issues. ☹️ 

We are likely to do a few semver majors in the nearish future to try to move from `import Foundation` to `import FoundationEssentials` where possible, but as you point out we can't do it for NIOFoundationCompat.

adam-fowler:
> We are likely to do a few semver majors in the nearish future to try to move from import Foundation to import FoundationEssentials where possible, but as you point out we can't do it for NIOFoundationCompat.

There could be an argument for deprecating the JSONSerialization support.

PR #212, opened 2024-10-17
Provide configurability for receiving connection data

clintonpi:
Motivation:

Users are stuck with the hardcoded parameters for receiving data from a connection.

Modifications:

- Add new options to `NIOTSChannelOptions` for configuring how to receive data from a connection.

Result:

Users can configure how they receive data from a connection.


Lukasa:
@swift-server-bot add to allowlist

Lukasa (approved):
Nice one! This looks great.

PR #230, opened 2025-04-09
Add `NWParameters` configurator to bootstraps

gjcairo:
Allow users to provide a closure taking an `NWParameters` to customise them before they're used to create `NWConnection`s.

### Motivation:

`NWParameters` are currently created using the provided TLS and UDP options, and then passed over to new `NWConnection`s. However, there are more ways in which `NWParameters` can be customised, so this new API provides a way for users to do this.

### Modifications:

Introduce new `configureNWParameters` methods to the existing bootstraps to allow configuring a closure for customising `NWParameters`.

### Result:

Users can now customise the `NWParameters` used to create new `NWConnection`s.

### DIFF
--- a/Sources/NIOTransportServices/Datagram/NIOTSDatagramListener.swift
+++ b/Sources/NIOTransportServices/Datagram/NIOTSDatagramListener.swift
@@ -64,9 +64,12 @@ public final class NIOTSDatagramListenerBootstrap {
     private var serverQoS: DispatchQoS?
     private var childQoS: DispatchQoS?
     private var udpOptions: NWProtocolUDP.Options = .init()
+    private var childUDPOptions: NWProtocolUDP.Options = .init()
     private var tlsOptions: NWProtocolTLS.Options?
+    private var childTLSOptions: NWProtocolTLS.Options?
     private var bindTimeout: TimeAmount?
     private var nwParametersConfigurator: (@Sendable (NWParameters) -> Void)?
+    private var childNWParametersConfigurator: (@Sendable (NWParameters) -> Void)?
 
     /// Create a ``NIOTSListenerBootstrap`` for the `EventLoopGroup` `group`.
     ///
@@ -225,23 +228,43 @@ public final class NIOTSDatagramListenerBootstrap {
         return self
     }
 
-    /// Specifies the TCP options to use on the child `Channel`s.
+    /// Specifies the TCP options to use on the listener.
     public func udpOptions(_ options: NWProtocolUDP.Options) -> Self {
         self.udpOptions = options
         return self
     }
 
-    /// Specifies the TLS options to use on the child `Channel`s.
+    /// Specifies the TCP options to use on the child `Channel`s.
+    public func childUDPOptions(_ options: NWProtocolUDP.Options) -> Self {
+        self.childUDPOptions = options
+        return self
+    }
+
+    /// Specifies the TLS options to use on the listener.
     public func tlsOptions(_ options: NWProtocolTLS.Options) -> Self {
         self.tlsOptions = options
         return self
     }
 
-    /// Customise the `NWParameters` to be used when creating the connection.
+    /// Specifies the TLS options to use on the child `Channel`s.
+    public func childTLSOptions(_ options: NWProtocolTLS.Options) -> Self {
+        self.childTLSOptions = options
+        return self
+    }
+
+    /// Customise the `NWParameters` to be used when creating the `NWConnection` for the listener.
     public func configureNWParameters(
         _ configurator: @Sendable @escaping (NWParameters) -> Void
     ) -> Self {
         self.nwParametersConfigurator = configurator
+        return self
+    }
+
+    /// Customise the `NWParameters` to be used when creating the `NWConnection`s for the child `Channel`s.
+    public func configureChildNWParameters(
+        _ configurator: @Sendable @escaping (NWParameters) -> Void
+    ) -> Self {
+        self.childNWParametersConfigurator = configurator
         return self
     }
 
@@ -339,9 +362,9 @@ public final class NIOTSDatagramListenerBootstrap {
                 nwParametersConfigurator: self.nwParametersConfigurator,
                 childLoopGroup: self.childGroup,
                 childChannelQoS: self.childQoS,
-                childUDPOptions: self.udpOptions,
-                childTLSOptions: self.tlsOptions,
-                childNWParametersConfigurator: self.nwParametersConfigurator
+                childUDPOptions: self.childUDPOptions,
+                childTLSOptions: self.childTLSOptions,
+                childNWParametersConfigurator: self.childNWParametersConfigurator
             )
         } else {
             serverChannel = NIOTSDatagramListenerChannel(
@@ -352,9 +375,9 @@ public final class NIOTSDatagramListenerBootstrap {
                 nwParametersConfigurator: self.nwParametersConfigurator,
                 childLoopGroup: self.childGroup,
                 childChannelQoS: self.childQoS,
-                childUDPOptions: self.udpOptions,
-                childTLSOptions: self.tlsOptions,
-                childNWParametersConfigurator: self.nwParametersConfigurator
+                childUDPOptions: self.childUDPOptions,
+                childTLSOptions: self.childTLSOptions,
+                childNWParametersConfigurator: self.childNWParametersConfigurator
             )
         }
 
--- a/Sources/NIOTransportServices/Datagram/NIOTSDatagramListener.swift
+++ b/Sources/NIOTransportServices/Datagram/NIOTSDatagramListener.swift
@@ -66,6 +66,7 @@ public final class NIOTSDatagramListenerBootstrap {
     private var udpOptions: NWProtocolUDP.Options = .init()
     private var tlsOptions: NWProtocolTLS.Options?
     private var bindTimeout: TimeAmount?
+    private var nwParametersConfigurator: (@Sendable (inout NWParameters) -> Void)?
 
     /// Create a ``NIOTSListenerBootstrap`` for the `EventLoopGroup` `group`.
     ///
@@ -236,6 +237,14 @@ public final class NIOTSDatagramListenerBootstrap {
         return self
     }
 
+    /// Customise the `NWParameters` to be used when creating the connection.
+    public func configureNWParameters(
+        _ configurator: @Sendable @escaping (inout NWParameters) -> Void
+    ) -> Self {
+        self.nwParametersConfigurator = configurator
+        return self
+    }
+
     /// Bind the `NIOTSListenerChannel` to `host` and `port`.
     ///
     /// - parameters:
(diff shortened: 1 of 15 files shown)


glbrntt (commented):
I think this shape is appropriate. Can you add some tests to verify that the configurator is actually called when expected? (And the appropriate number of times.)

gjcairo:
I added the tests and realised I was missing a couple of things + I can get rid of `inout` because `NWParameters` is a class.

Tests/NIOTransportServicesTests/NIOTSBootstrapTests.swift
```diff
@@ -371,6 +371,36 @@ final class NIOTSBootstrapTests: XCTestCase {
...
+            .configureNWParameters { _ in
+                configuratorConnectionCounter.withLockedValue { $0 += 1 }
+            }
+            .connect(to: listenerChannel.localAddress!)
+            .wait()
+
+        try listenerChannel.close().wait()
+        try connectionChannel.close().wait()
+
+        XCTAssertEqual(1, configuratorListenerCounter.withLockedValue { $0 })
```
glbrntt, line 401: Given the configurator is currently shared between the listening channel and accepted server channels, shouldn't the count here be two? (one for the listener, one for the accepted channel)

Sources/NIOTransportServices/Datagram/NIOTSDatagramListener.swift
```diff
@@ -327,21 +336,25 @@ public final class NIOTSDatagramListenerBootstrap {
                 qos: self.serverQoS,
                 udpOptions: self.udpOptions,
                 tlsOptions: self.tlsOptions,
+                nwParametersConfigurator: self.nwParametersConfigurator,
                 childLoopGroup: self.childGroup,
                 childChannelQoS: self.childQoS,
                 childUDPOptions: self.udpOptions,
-                childTLSOptions: self.tlsOptions
+                childTLSOptions: self.tlsOptions,
+                childNWParametersConfigurator: self.nwParametersConfigurator
```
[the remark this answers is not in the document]
gjcairo: The reason why I didn't have a separate one is because we don't make a distinction between TLS options and UDP options between child and parent - and we create the `NWParameters` for both child and parent from these same TLS and UDP options. 

But I suppose you may want to have _other_ different connection parameters/settings on the listener channel vs each child channel / connection?

Tests/NIOTransportServicesTests/NIOTSBootstrapTests.swift
gjcairo: There are separate counters for each configurator in the tests, so they will be 1 each.

The listener channel will create connection channels and pass down the configurator, but that happens when the `newConnectionHandler` is called by the underlying `NWListener`, which I'm not doing in this test.

Sources/NIOTransportServices/Datagram/NIOTSDatagramListener.swift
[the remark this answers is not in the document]
glbrntt: Oh I see, that makes sense why you didn't make the distinction, I do think more configurability is probably better here though. WDYT @Lukasa -- should we configure NWParameters separately for parent and child channels?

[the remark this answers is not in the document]
glbrntt: Sorry, latent thought: do they even apply to accepted channels?

Tests/NIOTransportServicesTests/NIOTSBootstrapTests.swift
glbrntt: `newConnectionHandler` should be getting called:
- You start a server which listens for connections (+1 listener counter)
- You start a client which opens a connection to the server (+1 connection counter)
- The server accepts that new connection and `newConnectionHandler` should be called so we'd expect a +1 listener counter.

Tests/NIOTransportServicesTests/NIOTSBootstrapTests.swift
```diff
@@ -18,7 +18,7 @@ import XCTest
 import Network
 import NIOCore
 import NIOEmbedded
-import NIOTransportServices
+@testable import NIOTransportServices
```
glbrntt, line 21: Why do we need the `@testable` here?

Tests/NIOTransportServicesTests/NIOTSBootstrapTests.swift
```diff
@@ -371,6 +371,60 @@ final class NIOTSBootstrapTests: XCTestCase {
...
+            connectionPromise: group.next().makePromise()
+        )
+
+        let listenerChannel = try await NIOTSListenerBootstrap(group: group)
+            .childChannelInitializer { connectionChannel in
+                connectionChannel.eventLoop.makeCompletedFuture {
+                    try connectionChannel.pipeline.syncOperations.addHandler(waitForConnectionHandler)
+                }
+            }
+            .configureNWParameters { _ in
```
glbrntt, line 404: Did we decide to not have a separate configurator for child channels?

Tests/NIOTransportServicesTests/NIOTSBootstrapTests.swift
```diff
@@ -371,6 +371,60 @@ final class NIOTSBootstrapTests: XCTestCase {
...
+        // Wait for the server to activate the connection channel to the client.
+        try await waitForConnectionHandler.connectionPromise.futureResult.get()
+
+        try await listenerChannel.close().get()
+        try await connectionChannel.close().get()
+
+        XCTAssertEqual(2, configuratorListenerCounter.withLockedValue { $0 })
+        XCTAssertEqual(1, configuratorConnectionCounter.withLockedValue { $0 })
+
+        try await group.shutdownGracefully()
```
glbrntt, line 426: This should ideally be done in a `defer` when created otherwise we might not shut it down if the test throws

Tests/NIOTransportServicesTests/NIOTSBootstrapTests.swift
```diff
@@ -371,6 +371,60 @@ final class NIOTSBootstrapTests: XCTestCase {
...
+            typealias InboundIn = Never
+
+            let connectionPromise: EventLoopPromise<Void>
+
+            init(connectionPromise: EventLoopPromise<Void>) {
+                self.connectionPromise = connectionPromise
+            }
+
+            func channelActive(context: ChannelHandlerContext) {
+                self.connectionPromise.succeed()
```
glbrntt, line 386: It won't matter in this test, but generally speaking handlers should forward handler lifecycle events. If they don't, any other handlers in the pipeline waiting for the event might not work as expected.

Tests/NIOTransportServicesTests/NIOTSBootstrapTests.swift
gjcairo: Ah, we don't, forgot to get rid of it when I added the test handler.

Sources/NIOTransportServices/Datagram/NIOTSDatagramListener.swift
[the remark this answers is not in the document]
gjcairo: Let's see what Cory's opinion is, but I'd go for either not allowing to have a separate configurator for child channels, or also allow users to provide child TLS and UDP options. I know they could use the closure to override the options but it's a weird API shape IMO to provide one but not the others.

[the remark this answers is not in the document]
Lukasa: I'm inclined to say we should do all of the above. And FWIW, users _can_ provide child TLS and UDP options, I believe: we just don't currently do anything with them.

Tests/NIOTransportServicesTests/NIOTSBootstrapTests.swift
```diff
@@ -371,6 +371,87 @@ final class NIOTSBootstrapTests: XCTestCase {
...
+                .configureNWParameters { _ in
+                    configuratorConnectionCounter.withLockedValue { $0 += 1 }
+                }
+                .connect(to: listenerChannel.localAddress!)
+                .get()
+
+            // Need to write something so the server can activate the connection channel: this is UDP,
+            // so there is no handshaking that happens and thus the server cannot know that the
+            // connection has been established and the channel can be activated until we receive something.
+            try await connectionChannel.writeAndFlush(ByteBuffer(bytes: [42]))
```
glbrntt, line 443: Copy-pasta: this isn't UDP, it's TCP

Tests/NIOTransportServicesTests/NIOTSBootstrapTests.swift
```diff
@@ -371,6 +371,87 @@ final class NIOTSBootstrapTests: XCTestCase {
...
+
+            try await listenerChannel.close().get()
+            try await connectionChannel.close().get()
+
+            XCTAssertEqual(1, configuratorListenerCounter.withLockedValue { $0 })
+            XCTAssertEqual(1, configuratorConnectionCounter.withLockedValue { $0 })
+        }
+    }
+
+    func testNWParametersConfigurator_ClientUsesConfigurator() async throws {
```
glbrntt, line 413: Is there any reason we shouldn't just test the client in the test above as well? The code is effectively the same. Or separate out the common logic into a helper  method. Either way we should de-duplicate this.

Tests/NIOTransportServicesTests/NIOTSDatagramConnectionChannelTests.swift
```diff
@@ -232,6 +233,92 @@ final class NIOTSDatagramConnectionChannelTests: XCTestCase {
...
+
+            try await listenerChannel.close().get()
+            try await connectionChannel.close().get()
+
+            XCTAssertEqual(1, configuratorListenerCounter.withLockedValue { $0 })
+            XCTAssertEqual(1, configuratorConnectionCounter.withLockedValue { $0 })
+        }
+    }
+
+    func testNWParametersConfigurator_ClientUsesConfigurator() async throws {
```
glbrntt, line 279: Similar question to above, should we just deduplicate this test?

Tests/NIOTransportServicesTests/NIOTSTestUtilities.swift
```diff
@@ -0,0 +1,43 @@
...
+//===----------------------------------------------------------------------===//
+
+#if canImport(Network)
+import NIOCore
+import NIOTransportServices
+
+func withEventLoopGroup(_ test: (EventLoopGroup) async throws -> Void) async rethrows {
+    let group = NIOTSEventLoopGroup()
+    do {
+        try await test(group)
```
glbrntt, line 22: You're not shutting down the ELG if `test` doesn't throw.

glbrntt (changes requested):

Tests/NIOTransportServicesTests/NIOTSBootstrapTests.swift
gjcairo: Hm yeah I was trying to make it clear that we were testing the client in one and the server on the other, but I merged them and renamed the counters and I think it's clear enough.

glbrntt (approved):
LGTM, thanks @gjcairo!

PR #233, opened 2025-04-16
Change naming of datagram-related types to make them more consistent + fix docs

gjcairo:
### Motivation:

The datagram-related types logically mirror the TCP-based ones (`NIOTSDatagramListenerBootstrap` vs `NIOTSListenerBootstrap`;  `NIOTSListenerChannel` vs `NIOTSDatagramListenerChannel`; `NIOTSDatagramChannel` vs `NIOTSDatagramConnectionChannel`; `NIOTSDatagramBootstrap` vs `NIOTSConnectionBootstrap`).
However, some of the type names could more closely resemble their TCP counterparts to make it easier to navigate the code/understand what their purpose is.
Additionally, the docs for some of these types are wrong, as they've been copy-pasted into the datagram versions without changes.

### Modifications:

There are separate commits for each of the following changes:
- Fix docs for `NIOTSDatagramListenerBootstrap` and rename the file to match the type name.
- Rename `NIOTSDatagramConnectionChannelTests` to `NIOTSDatagramBootstrapTests` (to match `NIOTSBootstrapTests`).
- Rename `NIOTSDatagramChannel` to `NIOTSDatagramConnectionChannel` (to match `NIOTSConnectionChannel`)
- Fix docs for `NIOTSConnectionBootstrap`.
- Rename `NIOTSDatagramBootstrap` to `NIOTSDatagramConnectionBootstrap` (to match `NIOTSConnectionBootstrap`). This one required a deprecate and replace since it's a public type.

### Result:

Better docs and more consistency in our type names.

### DIFF
--- a/Sources/NIOTransportServices/Datagram/NIOTSDatagramConnectionBootstrap.swift
+++ b/Sources/NIOTransportServices/Datagram/NIOTSDatagramConnectionBootstrap.swift
@@ -20,9 +20,10 @@ import Network
 @available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
 public typealias NIOTSDatagramBootstrap = NIOTSDatagramConnectionBootstrap
 
-/// A `NIOTSDatagramConnectionBootstrap` is an easy way to bootstrap a `NIOTSDatagramConnectionChannel` when creating network clients.
+/// A ``NIOTSDatagramConnectionBootstrap`` is an easy way to bootstrap a UDP channel when creating network clients.
 ///
-/// Usually you re-use a `NIOTSDatagramConnectionBootstrap` once you set it up, calling `connect` multiple times on the same bootstrap.
+/// Usually you re-use a ``NIOTSDatagramConnectionBootstrap`` once you set it up, calling `connect` multiple times on the
+/// same bootstrap.
 /// This way you ensure that the same `EventLoop`s will be shared across all your connections.
 ///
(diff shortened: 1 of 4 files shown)


Sources/NIOTransportServices/Datagram/NIOTSDatagramConnectionBootstrap.swift
```diff
@@ -17,9 +17,13 @@ import NIOCore
 import Dispatch
 import Network
 
-/// A `NIOTSDatagramBootstrap` is an easy way to bootstrap a `NIOTSDatagramChannel` when creating network clients.
+@available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
+@available(*, deprecated, renamed: "NIOTSDatagramConnectionBootstrap")
+public typealias NIOTSDatagramBootstrap = NIOTSDatagramConnectionBootstrap
```
gjcairo, line 22: I understand if we may want to avoid this - I can roll this change back if we don't want to deprecate the existing type.

gjcairo: Sure, if we're okay with the type alias then I'd vote for that but without deprecating.

glbrntt (approved):
Looks fine but want to check @Lukasa is happy with renaming a bunch of stuff

Lukasa (approved):

PR #236, opened 2025-05-08
Added buffer pool to NIOTSConnectionChannel

rafaelcepeda:
Added buffer pool to NIOTSConnectionChannel for reading messages.

### Motivation:

In order to avoid creating a new `ByteBuffer` for every single message read from the channel, we decided to reuse the `NIOPooledRecvBufferAllocator` type from NIOCore to leverage a pool of buffers.

### Modifications:

Used the buffer allocation mechanism provided by the `NIOPooledRecvBufferAllocator` to reuse and adapt previously created receive buffers.

### Result:

Channel reads can now reuse previously created buffers, avoiding unnecessary overhead by creating new buffers every single time.

### DIFF
--- a/Sources/NIOTransportServices/NIOTSConnectionChannel.swift
+++ b/Sources/NIOTransportServices/NIOTSConnectionChannel.swift
@@ -129,7 +129,7 @@ internal struct BackpressureManager {
 @available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
 internal final class NIOTSConnectionChannel: StateManagedNWConnectionChannel {
     /// The `ByteBufferAllocator` for this `Channel`.
-    public let allocator = ByteBufferAllocator()
+    public var allocator = ByteBufferAllocator()
 
     /// An `EventLoopFuture` that will complete when this channel is finally closed.
     public var closeFuture: EventLoopFuture<Void> {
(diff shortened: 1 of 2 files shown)


glbrntt (commented):
LGTM aside from one nit. We'll also need to do a NIO release and bump the required version before we can merge this.

Sources/NIOTransportServices/NIOTSConnectionChannel.swift
```diff
@@ -236,6 +236,15 @@ internal final class NIOTSConnectionChannel: StateManagedNWConnectionChannel {
     /// A lock that guards the _addressCache.
     internal let _addressCacheLock = NIOLock()
 
+    /// The `NIOPooledRecvBufferAllocator` used to allocate buffers for incoming data
+    internal var recvBufferPool: NIOPooledRecvBufferAllocator
```
rafaelcepeda, line 240: @glbrntt should I rename this property to `_ recvBufferPool`?

Sources/NIOTransportServices/NIOTSConnectionChannel.swift
```diff
@@ -129,7 +129,7 @@ internal struct BackpressureManager {
 @available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
 internal final class NIOTSConnectionChannel: StateManagedNWConnectionChannel {
     /// The `ByteBufferAllocator` for this `Channel`.
-    public let allocator = ByteBufferAllocator()
+    public var allocator = ByteBufferAllocator()
```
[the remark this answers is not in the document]
rafaelcepeda: Yeah, I think this is a spillover from a previous attempt. Just changed it back to be a constant.

Sources/NIOTransportServices/NIOTSConnectionChannel.swift
glbrntt: Shouldn't it be `private` rather than `internal`?

It should only be prefixed with an underscore if it was meant to be `private` but can't be because it needs to be `@usableFromInline`.

PR #243, opened 2025-09-10
Fix receiving more than 1 blob of data on a NWConnection

Joannis:
Prevent closing the channel on `isFinished` for a datagram message

### Motivation:

NWConnection for UDP currently closes too soon, it closes after 1 completed read.

### Modifications:

Don't close the connection on UDP read.

### DIFF
--- a/Sources/NIOTransportServices/StateManagedNWConnectionChannel.swift
+++ b/Sources/NIOTransportServices/StateManagedNWConnectionChannel.swift
@@ -48,6 +48,9 @@ internal protocol StateManagedNWConnectionChannel: StateManagedChannel where Act
     var nwOptions: NWOptions { get }
 
     var connection: NWConnection? { get set }
+    
+    /// If `true`, `isComplete` on a NWConnection's `receive` callback indicates the socket has closed
+    var completionClosesChannel: Bool { get }
 
     var minimumIncompleteReceiveLength: Int { get set }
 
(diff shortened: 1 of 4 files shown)


Sources/NIOTransportServices/StateManagedNWConnectionChannel.swift
```diff
@@ -438,15 +441,20 @@ extension StateManagedNWConnectionChannel {
             var buffer = self.allocator.buffer(capacity: content.count)
             buffer.writeBytes(content)
             self.pipeline.fireChannelRead(buffer)
-            self.pipeline.fireChannelReadComplete()
+            
+            if isComplete {
+                self.pipeline.fireChannelReadComplete()
+            } else if !completionClosesChannel {
+                self.pipeline.fireChannelReadComplete()
+            }
```
Lukasa, line 449: Perhaps I'm missing something: why is this change here? We now don't `fireChannelReadComplete` if `isComplete == false` and `completionClosesChannel == true`: what's the motivation?

camunro: you may also want to consider adding a default to StateManagedNWConnectionChannel and not need to touch NIOTSConnectionChannel.swift :

```
extension StateManagedNWConnectionChannel {
    public var completionClosesChannel: Bool { true }
}
```

acrqualityfiber-oss (approved):

PR #244, opened 2025-09-15
Fixes premature closure of NIOTS UDP channels by not treating per-datagram isComplete as EOF.

camunro:
Fix NIOTS UDP premature close by not treating per‑datagram isComplete as EOF; add multi‑datagram tests and NIOTSEchoClient example.

  Motivation:

  - While using [DNSClient #40](https://github.com/orlandos-nl/DNSClient/issues/40), we observed “one‑shot” behavior: a NIOTS connected‑UDP channel closed after the first datagram, causing subsequent writes to fail or be dropped.
  - Network.framework sets isComplete == true for each UDP datagram. This indicates message completeness, not stream EOF. Treating that signal as EOF caused the early close.
  - [PR#243](https://github.com/apple/swift-nio-transport-services/pull/243) explored gating behavior differently (e.g., via “completionClosesChannel”). This PR aligns EOF handling explicitly to channel semantics (stream vs. datagram), which makes the reasoning around transport semantics clear.
  - Alternatives considered (brief):
      - ContentContext.isFinal only: common stream half‑closes (FIN) arrive with isComplete == true and isFinal == false, so relying solely on isFinal would miss EOF on streams and risk hangs.
      - protocolMetadata (e.g., NWProtocolUDP.Metadata): metadata presence/shape varies and is often empty on final callbacks; channel semantics (isDatagramChannel) are explicit and robust.
  - Reproducer (pre‑fix): set a symbolic breakpoint `br s -n nw_connection_cancel`; run `testBasicChannelCommunication()` test; typical backtrace after the echoed datagram:
      - StateManagedNWConnectionChannel.dataReceivedHandler(content=<N bytes>, isComplete=true, error=nil) → didReadEOF() → StateManagedChannel.close0(error=eof) → nw_connection_cancel.
      - Note error=nil: close was triggered solely because isComplete was true.

  Modifications:

  - Transport fix:
      - StateManagedNWConnectionChannel.dataReceivedHandler now calls didReadEOF() only when isComplete && !isDatagramChannel (streams unchanged; datagrams no longer misinterpret isComplete as EOF).
  - Tests:
      - NIOTSDatagramBootstrapTests.testConnectedUDPEchoesTwoDatagrams verifies two echoes on NIOTS connected UDP; both server and client observe datagrams; bounded timeouts; proper cleanup.
      - NIOTSUDPMultiDatagramsPOSIXServerTests.testNIOTSClient_POSIXServer_MultipleDatagrams verifies NIOTS client ↔ POSIX UDP echo; asserts two echoes to isolate cross‑stack behavior.
  - Example:
      - NIOTSEchoClient executable example (Apple platforms): simple connected‑UDP echo client using NIOTSDatagramConnectionBootstrap; guarded with `canImport(Network)` and `@available`; minimal usage notes.

  Result:

  - Datagram channels (UDP) no longer misinterpret per‑packet completeness as EOF; multi‑datagram send/receive works as expected.
  - Stream channels (TCP/TLS) are unchanged; EOF continues to be detected per existing behavior.
  - Restores expected behavior so typical NIO UDP apps can be ported to NIOTS by swapping event loops and datagram bootstraps (on Apple platforms using Network.framework).
  - Manual validation available via NIOTSEchoClient (pair with SwiftNIO’s NIOUDPEchoServer).

### DIFF
--- a/Sources/NIOTSEchoClient/NIOTSEchoClient.swift
+++ b/Sources/NIOTSEchoClient/NIOTSEchoClient.swift
@@ -31,21 +31,21 @@ import Darwin
 final class EchoHandler: ChannelInboundHandler {
     typealias InboundIn = ByteBuffer
     typealias OutboundOut = ByteBuffer
-    
+
     func channelActive(context: ChannelHandlerContext) {
         print("Channel active")
     }
-    
+
     func channelRead(context: ChannelHandlerContext, data: NIOAny) {
         var buf = self.unwrapInboundIn(data)
         let text = buf.readString(length: buf.readableBytes) ?? "<\(buf.readableBytes) bytes>"
         print("echo:", text)
     }
-    
+
     func channelInactive(context: ChannelHandlerContext) {
         print("Channel inactive (closed)")
     }
-    
+
     func errorCaught(context: ChannelHandlerContext, error: Error) {
         print("error:", error)
         // Intentionally do not close here; keep running to observe behavior.
--- a/Sources/NIOTSEchoClient/NIOTSEchoClient.swift
+++ b/Sources/NIOTSEchoClient/NIOTSEchoClient.swift
@@ -1,12 +1,12 @@
 //===----------------------------------------------------------------------===//
 //
-// This source file is part of the NIOTSEchoClient open source project
+// This source file is part of the SwiftNIO open source project
 //
-// Copyright (c) 2025 Route Objects and the NIOTSEchoClient project authors
+// Copyright (c) 2017-2025 Apple Inc. and the SwiftNIO project authors
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information
-// See CONTRIBUTORS.txt for the list of NIOTSEchoClient project authors
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
 //
 // SPDX-License-Identifier: Apache-2.0
 //
@@ -27,7 +27,7 @@ import NIOCore
 import NIOTransportServices
 import Darwin
 
-@available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
+@available(macOS 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
 final class EchoHandler: ChannelInboundHandler {
     typealias InboundIn = ByteBuffer
     typealias OutboundOut = ByteBuffer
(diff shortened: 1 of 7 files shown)


Lukasa:
Thanks for this! There is already a PR open for this: #243. @Joannis do you want to coordinate with @camunro to figure out which of these two should go forward?

Joannis:
We'll do @Lukasa

camunro:
Hi @Lukasa,

Thanks for the update. I've connected with @Joannis, and he has graciously suggested that we move forward with my pull request.

We are in agreement to proceed with PR #244. Please let me know if there's anything else needed from my end.

Thanks

Sources/NIOTransportServices/StateManagedNWConnectionChannel.swift
```diff
@@ -100,6 +104,8 @@ extension StateManagedNWConnectionChannel {
         self
     }
 
+    public var isDatagramChannel: Bool { false }
```
Lukasa, line 107: This can avoid being `public`, as this field isn't part of a public protocol so its visibility won't actually be `public` anyway.

Sources/NIOTSEchoClient/NIOTSEchoClient.swift
```diff
@@ -0,0 +1,111 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the NIOTSEchoClient open source project
+//
+// Copyright (c) 2025 Route Objects and the NIOTSEchoClient project authors
```
Lukasa, line 5: I'm afraid the header file does need to exactly match the others.

Sources/NIOTSEchoClient/NIOTSEchoClient.swift
```diff
@@ -0,0 +1,111 @@
...
+//   swift run NIOUDPEchoServer 127.0.0.1 9999
+//
+
+#if canImport(Network)
+
+import NIOCore
+import NIOTransportServices
+import Darwin
+
+@available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
```
Lukasa, line 30:
```suggestion
@available(macOS 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
```

Lukasa (approved):

Lukasa (approved):

